Issue #2, opened 2019-08-28
Provide an UI for Gerrit to configure Eiffel Message sending enabled and REMReM configurations


SantoshNC68:
Dependent on #8 

Investigate what configurations the GUI should provide with regards to the architecture overview seen here #6. Some configurations that should exist at minimum is:
- The enabling of event sending
- REMReM configuration
- Selection of which events you want to trigger on

magnusbaeck:
Are you targeting PolyGerrit, the legacy GWT UI, or both?

SantoshNC68:
At present it will be Legacy GWT UI then the changes will be added on top if it is non compatible.

PR #15, opened 2019-09-24
Added configurations for Gerrit Plugin in UI

SantoshNC68:
### Applicable Issues
#2 

### Description of the Change
Added configuration part for the Gerrit plugin.

### Sign-off

    indicated in the file; or

    in the file; or

    it.

Signed-off-by:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -51,6 +51,20 @@
             <version>1.7.26</version>
             <scope>test</scope>
         </dependency>
+        
+        <!-- Added for Unit testing the UI -->
+        <dependency>
+            <groupId>org.assertj</groupId>
+            <artifactId>assertj-core</artifactId>
+            <version>3.9.1</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.powermock</groupId>
+            <artifactId>powermock-api-mockito</artifactId>
+            <version>1.7.4</version>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 
     <build>
--- a/pom.xml
+++ b/pom.xml
@@ -7,12 +7,12 @@
     <groupId>eiffel-community</groupId>
     <artifactId>eiffel-gerrit-plugin</artifactId>
     <packaging>jar</packaging>
-    <version>0.0.1-SNAPSHOT</version>
+    <version>0.0.1</version>
 
     <properties>
         <Gerrit-ApiType>plugin</Gerrit-ApiType>
         <Gerrit-ApiVersion>2.14.16</Gerrit-ApiVersion>
-        <Gerrit-PluginName>Eiffel-Agen</Gerrit-PluginName>
+        <Gerrit-PluginName>Eiffel-Integration</Gerrit-PluginName>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <skipTests>false</skipTests>
         <skipUTs>${skipTests}</skipUTs>
--- a/pom.xml
+++ b/pom.xml
@@ -54,12 +54,6 @@
         
         <!-- Added for Unit testing the UI -->
         <dependency>
-            <groupId>org.assertj</groupId>
-            <artifactId>assertj-core</artifactId>
-            <version>3.9.1</version>
-            <scope>test</scope>
-        </dependency>
-        <dependency>
             <groupId>org.powermock</groupId>
             <artifactId>powermock-api-mockito</artifactId>
             <version>1.7.4</version>
--- a/src/main/java/com/ericsson/gerrit/plugins/eiffel/GerritModule.java
+++ b/src/main/java/com/ericsson/gerrit/plugins/eiffel/GerritModule.java
@@ -54,8 +54,6 @@ public class GerritModule extends AbstractModule {
                 .toInstance(new ProjectConfigEntry("Filter branch", ""));
         bind(ProjectConfigEntry.class).annotatedWith(Exports.named(EiffelPluginConfiguration.FLOW_CONTEXT))
                 .toInstance(new ProjectConfigEntry("Flow Context", ""));
-        bind(ProjectConfigEntry.class).annotatedWith(Exports.named(EiffelPluginConfiguration.REMREM_GENERATE_URL))
-                .toInstance(new ProjectConfigEntry("REMReM Generate URL", ""));
         bind(ProjectConfigEntry.class).annotatedWith(Exports.named(EiffelPluginConfiguration.REMREM_PUBLISH_URL))
                 .toInstance(new ProjectConfigEntry("REMReM Publish URL", ""));
         bind(ProjectConfigEntry.class).annotatedWith(Exports.named(EiffelPluginConfiguration.REMREM_USERNAME))
--- a/src/main/java/com/ericsson/gerrit/plugins/eiffel/GerritModule.java
+++ b/src/main/java/com/ericsson/gerrit/plugins/eiffel/GerritModule.java
@@ -58,6 +58,8 @@ public class GerritModule extends AbstractModule {
                 .toInstance(new ProjectConfigEntry("REMReM Publish URL", ""));
         bind(ProjectConfigEntry.class).annotatedWith(Exports.named(EiffelPluginConfiguration.REMREM_USERNAME))
                 .toInstance(new ProjectConfigEntry("REMReM Username", ""));
+        
+        // Currently the Gerrit has defined set of types that can be used. The Password is String type today, but will need some changes.
         bind(ProjectConfigEntry.class).annotatedWith(Exports.named(EiffelPluginConfiguration.REMREM_PASSWORD))
                 .toInstance(new ProjectConfigEntry("REMReM Password", ""));
         // Register change listener that will send messages
--- a/src/main/java/com/ericsson/gerrit/plugins/eiffel/GerritModule.java
+++ b/src/main/java/com/ericsson/gerrit/plugins/eiffel/GerritModule.java
@@ -48,21 +48,20 @@ public class GerritModule extends AbstractModule {
                                      .to(MessageQueueHandler.class);
 
         // Example of how to register plugin configuration to the project screen
-         bind(ProjectConfigEntry.class).annotatedWith(Exports.named(EiffelPluginConfiguration.ENABLED))
-         .toInstance(new ProjectConfigEntry("Enable Eiffel messaging",
-         false));
-         bind(ProjectConfigEntry.class).annotatedWith(Exports.named(EiffelPluginConfiguration.FILTER))
-         .toInstance(new ProjectConfigEntry("Filter branch", ""));
-         bind(ProjectConfigEntry.class).annotatedWith(Exports.named(EiffelPluginConfiguration.FLOW_CONTEXT))
-         .toInstance(new ProjectConfigEntry("Flow Context", ""));
-         bind(ProjectConfigEntry.class).annotatedWith(Exports.named(EiffelPluginConfiguration.REMREM_GENERATE_URL))
-         .toInstance(new ProjectConfigEntry("REMReM Generate URL", ""));
-         bind(ProjectConfigEntry.class).annotatedWith(Exports.named(EiffelPluginConfiguration.REMREM_PUBLISH_URL))
-         .toInstance(new ProjectConfigEntry("REMReM Publish URL", ""));
-         bind(ProjectConfigEntry.class).annotatedWith(Exports.named(EiffelPluginConfiguration.REMREM_USERNAME))
-         .toInstance(new ProjectConfigEntry("REMReM Username", ""));
-         bind(ProjectConfigEntry.class).annotatedWith(Exports.named(EiffelPluginConfiguration.REMREM_PASSWORD))
-         .toInstance(new ProjectConfigEntry("REMReM Password", ""));
+        bind(ProjectConfigEntry.class).annotatedWith(Exports.named(EiffelPluginConfiguration.ENABLED))
+                .toInstance(new ProjectConfigEntry("Enable Eiffel messaging", false));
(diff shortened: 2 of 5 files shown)


pom.xml
```diff
@@ -51,6 +51,20 @@
             <version>1.7.26</version>
             <scope>test</scope>
         </dependency>
+        
+        <!-- Added for Unit testing the UI -->
+        <dependency>
+            <groupId>org.assertj</groupId>
```
a-breid-ericsson: This dependency may not be needed if you use:
org.junit.Assert* 
 *assertEquals
 *assertTrue
 *assertFalse

Less dependencies used the better :)

src/main/java/com/ericsson/gerrit/plugins/eiffel/configuration/EiffelPluginConfiguration.java
```diff
@@ -46,7 +64,70 @@ public EiffelPluginConfiguration(final String pluginName, final NameKey project,
...
+                            project.toString()));
+        } else if (this.cfgRemremUsername == null) {
+            throw new ExceptionInInitializerError(
+                    String.format("Can't read %s plugin configuration for project %s: REMReM Username is null", pluginName,
+                            project.toString()));
+        } else if (this.cfgRemremPassword == null) {
+            throw new ExceptionInInitializerError(
+                    String.format("Can't read %s plugin configuration for project %s: REMReM Password is null", pluginName,
+                            project.toString()));
+        }
```
m-linner-ericsson: There is a lot of code duplication. Is it possible to create a method that checks and throws on null?

SantoshNC68: this is needed as we need to show proper error messages on what property is missing.

src/main/resources/Documentation/about.md
```diff
@@ -8,11 +8,44 @@ Parameters
...
+eiffel-agen.enabled
+:   Plugin state. if set to `true`, send message Eiffel message for every commit submitted for this repository.
+    Plugin considered disabled if this option is not defined or set to `false`.
+
+eiffel-agen.filter
+:   Allow users to defined a white list of branches, i.e. messages will be sent only for a commits submitted on specified branches, all other submits will be ignored.
+    Accepts Java regex's separated by spaces. If not defined, send messages for commits submitted on all branches.
+
+eiffel-agen.flow-context
+:   Optional, but expects a UUID of a EiffelFlowContextDefinedEvent if provided.
```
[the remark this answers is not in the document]
SantoshNC68: It can be multiple events, they will link to the FLOW-CONTEXT link type.
In a CI flow you should be able to know which FLOW this events belong to. It can be a single or multiple Flows.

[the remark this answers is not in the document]
SantoshNC68: For the first iteration, we are expecting to provide the UUID, but later it will support lookups which will try to fetch the UUID and add them in the events.

src/main/java/com/ericsson/gerrit/plugins/eiffel/configuration/EiffelPluginConfiguration.java
m-linner-ericsson: I wasn't asking to remove it I was wondering if could be simplified e.g:

```
throwOnNull(plugin, project, "REMRem Generate URL is null");
throwOnNull(plugin, project, "REMRem Publish URL is null");
...
```

Currently it is a bit confusing why we have the `else if` as I guess we want to check all values?

src/main/java/com/ericsson/gerrit/plugins/eiffel/GerritModule.java
```diff
@@ -45,11 +48,18 @@ protected void configure() {
...
+                .toInstance(new ProjectConfigEntry("Enable Eiffel messaging", false));
+        bind(ProjectConfigEntry.class).annotatedWith(Exports.named(EiffelPluginConfiguration.FILTER))
+                .toInstance(new ProjectConfigEntry("Filter branch", ""));
+        bind(ProjectConfigEntry.class).annotatedWith(Exports.named(EiffelPluginConfiguration.FLOW_CONTEXT))
+                .toInstance(new ProjectConfigEntry("Flow Context", ""));
+        bind(ProjectConfigEntry.class).annotatedWith(Exports.named(EiffelPluginConfiguration.REMREM_PUBLISH_URL))
+                .toInstance(new ProjectConfigEntry("REMReM Publish URL", ""));
+        bind(ProjectConfigEntry.class).annotatedWith(Exports.named(EiffelPluginConfiguration.REMREM_USERNAME))
+                .toInstance(new ProjectConfigEntry("REMReM Username", ""));
+        bind(ProjectConfigEntry.class).annotatedWith(Exports.named(EiffelPluginConfiguration.REMREM_PASSWORD))
```
e-edling-ericsson: So the password is stored in plain text right now? And is this field visible in the GUI? If we go for this in the first iteration, I think you should create an issue immediately about changing it later.

src/main/java/com/ericsson/gerrit/plugins/eiffel/configuration/EiffelPluginConfiguration.java
e-edling-ericsson: I second that @m-linner-ericsson

src/main/java/com/ericsson/gerrit/plugins/eiffel/configuration/EiffelPluginConfiguration.java
```diff
@@ -46,7 +62,71 @@ public EiffelPluginConfiguration(final String pluginName, final NameKey project,
...
+        this.enabled = pluginConfig.getBoolean(ENABLED, false);
+        this.filter = pluginConfig.getString(FILTER);
+        this.remremPublishURL = pluginConfig.getString(REMREM_PUBLISH_URL);
+        this.remremUsername = pluginConfig.getString(REMREM_USERNAME);
+        this.remremPassword = pluginConfig.getString(REMREM_PASSWORD);
+        //flow context is optional
+        this.flowContext = pluginConfig.getString(FLOW_CONTEXT);
 
+        // No point to check other config parameters if plugin is disabled
+        boolean isValidated = validatePluginConfig(pluginName, project);
```
e-edling-ericsson: I think "isValidated" is the wrong word. I would probably say, isValid or isConfigValid to make it more clear. :)

src/main/resources/Documentation/about.md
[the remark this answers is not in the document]
e-pettersson-ericsson: Good, now I think it's more clear and the link to the Eiffel protocol repository helps guide users who want more information.

src/main/java/com/ericsson/gerrit/plugins/eiffel/configuration/EiffelPluginConfiguration.java
```diff
@@ -46,7 +62,71 @@ public EiffelPluginConfiguration(final String pluginName, final NameKey project,
             throw new ExceptionInInitializerError(String.format("Can't read %s plugin configuration for project %s: %s",
                     pluginName, project.toString(), e.getMessage()));
         }
```
e-edling-ericsson: Since this method is now a bit bigger, could you break the try catch code into a method(loadPluginConfig(project, pluginName)).

src/main/java/com/ericsson/gerrit/plugins/eiffel/configuration/EiffelPluginConfiguration.java
```diff
@@ -46,7 +62,71 @@ public EiffelPluginConfiguration(final String pluginName, final NameKey project,
...
+     * @param project
+     */
+    private boolean validatePluginConfig(final String pluginName, final NameKey project) {
+        if (!this.enabled) {
+            return false;
+        }
+        
+        // Make sure that REMReM configuration is defined, otherwise we won't be able to send messages.
+        if (this.remremPublishURL == null) {
+            throw new ExceptionInInitializerError(
```
e-edling-ericsson: Is this caught somewhere? Maybe it's not needed, i don't know exactly how the plugin works.

src/main/java/com/ericsson/gerrit/plugins/eiffel/configuration/EiffelPluginConfiguration.java
```diff
@@ -46,7 +62,71 @@ public EiffelPluginConfiguration(final String pluginName, final NameKey project,
             throw new ExceptionInInitializerError(String.format("Can't read %s plugin configuration for project %s: %s",
                     pluginName, project.toString(), e.getMessage()));
         }
+        // Read plugin configuration
+        this.enabled = pluginConfig.getBoolean(ENABLED, false);
+        this.filter = pluginConfig.getString(FILTER);
```
henning-roos: You could return directly here if plugin is not enabled, instead of inside the validation

src/main/java/com/ericsson/gerrit/plugins/eiffel/configuration/EiffelPluginConfiguration.java
henning-roos: If my above suggestion is followed, the boolean can be skipped. The throws will handle any exceptions from validate.

src/main/java/com/ericsson/gerrit/plugins/eiffel/configuration/EiffelPluginConfiguration.java
SantoshNC68: This cannot be done as the Constructor expects all the fields to have been initiated.

src/main/java/com/ericsson/gerrit/plugins/eiffel/configuration/EiffelPluginConfiguration.java
SantoshNC68: This is needed when the plugin is enabled and the configurations are not set. The plugin will not load and fail.

src/main/resources/Documentation/about.md
```diff
@@ -8,11 +8,45 @@ Parameters
 
 Project Configuration:
 
-TODO: Add Configuration parameters and instructions.
+eiffel-agen.enabled
```
tobiasake: Why do you have protocol release name/version in configuration?

src/main/java/com/ericsson/gerrit/plugins/eiffel/configuration/EiffelPluginConfiguration.java
m-linner-ericsson: That is put a 
```
if(!this.enabled){
return;
}
```

at line 67 instead of the if-statements on 76 and 88. It will be the same logic as the `validatePluginConfig` returns directly if the plugin is not enabled.

src/main/java/com/ericsson/gerrit/plugins/eiffel/GerritModule.java
SantoshNC68: Yes, it is plain text and there is no type that can say this field as a password. 

I added a comment on top of the field so that we need to modify it later.

src/main/java/com/ericsson/gerrit/plugins/eiffel/configuration/EiffelPluginConfiguration.java
SantoshNC68: I have removed most of the validations, and just kept the Publish URL as the mandatory one.

src/main/java/com/ericsson/gerrit/plugins/eiffel/configuration/EiffelPluginConfiguration.java
SantoshNC68: This has been moved. This will not start the plugin if it is empty.

src/main/java/com/ericsson/gerrit/plugins/eiffel/configuration/EiffelPluginConfiguration.java
SantoshNC68: Modified the validations, so I only have 1 check now.

src/main/resources/Documentation/about.md
SantoshNC68: Because that is how the plugin is built. In the pom.xml it has eiffel-agen, and when you edit the configurations using Gerrit UI, this is how it will look like.

src/main/java/com/ericsson/gerrit/plugins/eiffel/configuration/EiffelPluginConfiguration.java
m-linner-ericsson: Wasn't the other checks needed?

SantoshNC68: I removed, because now we only intend to use REMReM Publish, and the username password are optional when you configure REMReM.

SantoshNC68:
> Looks good.
> 
> Have you tested to build and load the plugin in Gerrit (in docker) ?
> Do that before merge.

Yes tested in docker environment and it seems to load the UI and configurations in gerrit.

src/main/resources/Documentation/about.md
tobiasake: So we need to changed parameter names for every Eiffel Protocol Release?
Why can't it be named only "eiffel.enabled" without any Eiffel version names ?

src/main/java/com/ericsson/gerrit/plugins/eiffel/configuration/EiffelPluginConfiguration.java
e-edling-ericsson: I didnt mean that the check isn't needed. I meant, is it needed to catch the Exception, because it was not cought somewhere. :) I understand the confusion though because i was not clear enough

src/main/resources/Documentation/about.md
tobiasake: ok, discussed and solved.

PR #21, opened 2019-09-30
Use copy not add

m-linner-ericsson:
### Description of the Change
- In the docker file use `COPY` instead of `ADD`
- Make a more meaningful name for the input parameter for the docker file

### Alternate Designs
Keep using `ADD`

### Benefits
- Docker themselves recommend using `COPY` instead of `ADD` https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#add-or-copy
- The `ADD` commands behaves differently depending on what is added. The rules in the documentation https://docs.docker.com/engine/reference/builder/#add are a bit tricky.  Example 

  > Note: Whether a file is identified as a recognized compression format or not is done solely based on the contents of the file, not the name of the file. For example, if an empty file happens to end with `.tar.gz` this **will not** be recognized as a compressed file and will not generate any kind of decompression error message, rather the file will simply be copied to the destination.

- `ADD` does not handle authentication (stated in the documentation)
- `ADD` will make the cache invalid, even for RUN commands (as stated in the documentation)
- As stated on https://dzone.com/articles/why-you-should-use-copy-instead-of-add-when-buildi there are a couple of security concerns to think of

### Possible Drawbacks
- If a user wants to take something from a URL they would have to download it first
- Compressed files are not decompressed

### Sign-off

    indicated in the file; or

    in the file; or

    it.

Signed-off-by: Mattias Linnér mattias.linner@ericsson.com

### DIFF
--- a/src/main/docker/Dockerfile
+++ b/src/main/docker/Dockerfile
@@ -19,7 +19,7 @@ USER gerrit
 RUN java -jar /var/gerrit/bin/gerrit.war init --batch -d /var/gerrit
 RUN java -jar /var/gerrit/bin/gerrit.war reindex -d /var/gerrit
 
-copy ${PLUGIN_LOCATION} /var/gerrit/plugins/eiffel-plugin.jar
+COPY ${PLUGIN_LOCATION} /var/gerrit/plugins/eiffel-plugin.jar
 
 # Allow incoming traffic
 EXPOSE 29418 8080
--- a/src/main/docker/Dockerfile
+++ b/src/main/docker/Dockerfile
@@ -2,7 +2,7 @@ FROM centos:7.5.1804
 
 LABEL maintainer="Eiffel-Community"
 
-ARG PLUGIN_LOCATION
+ARG JAR_LOCATION
 
 RUN yum -y install openssh-client initscripts sudo
 
@@ -19,7 +19,7 @@ USER gerrit
 RUN java -jar /var/gerrit/bin/gerrit.war init --batch -d /var/gerrit
 RUN java -jar /var/gerrit/bin/gerrit.war reindex -d /var/gerrit
 
-COPY ${PLUGIN_LOCATION} /var/gerrit/plugins/eiffel-plugin.jar
+COPY ${JAR_LOCATION} /var/gerrit/plugins/eiffel-plugin.jar
 
 # Allow incoming traffic
 EXPOSE 29418 8080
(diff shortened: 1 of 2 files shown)


src/main/docker/Dockerfile
```diff
@@ -19,7 +19,7 @@ USER gerrit
 RUN java -jar /var/gerrit/bin/gerrit.war init --batch -d /var/gerrit
 RUN java -jar /var/gerrit/bin/gerrit.war reindex -d /var/gerrit
 
-ADD ${URL} /var/gerrit/plugins/eiffel-plugin.jar
+COPY ${PLUGIN_LOCATION} /var/gerrit/plugins/eiffel-plugin.jar
```
tobiasake: How can we provide an URL address to artifactory where eiffel-plugin.jar is located when it is changed to COPY method?
I guess you can change to CMD  and using "curl" or "wget" or something instead, if you don't prefer ADD.

m-linner-ericsson: Yes a person will have to download the item first. `ADD` brings too many issues and vulnerabilities as described

src/main/docker/README.md
```diff
@@ -26,8 +26,8 @@ Build Eiffel Gerrit Plugin Docker image:
 
 ```bash
 cd (git root dir)
-export URL=target/$(ls target/ | grep '^eiffel-gerrit-plugin-[0-9]*.[0-9]*.[0-9]*.jar')
-docker build -t eiffel-gerrit-plugin --build-arg URL=./${URL} -f src/main/docker/Dockerfile .
+export PLUGIN_LOCATION=target/$(ls target/ | grep '^eiffel-gerrit-plugin-[0-9]*.[0-9]*.[0-9]*.jar')
```
e-pettersson-ericsson: I think one of the reasons we used URL before was so we could keep the build process generic for several Eiffel component Docker files, and we could then re-use the same automated script for building Docker images etc. 

This PLUGIN_LOCATION is very specific to this particular component, but I'm hesitant since I'm not sure if we need to keep it generic for all Eiffel components ... How about using JAR_LOCATION?

m-linner-ericsson: I think the cons outweigh the use. I have changed the argument in 2e9b0dc

src/main/docker/Dockerfile
e-edling-ericsson: I agree with @m-linner-ericsson in this case. The user have to download the item first.

tobiasake: Ok,, it is just more manual step when building and releasing the Docker images, especially when I going through all Eiffel components and build all latest released versions from artifactories.
If we agree on this way, then I am OK with the change.
I can mention in docker-compose files we have in several Eiffel components,uses some functionality that has been deprecated and removed in later version of docker-compose, e.g. the "health-check" function.

PR #28, opened 2019-10-15
Add REMReM sender class

### LINKED ISSUE #7: The Gerrit Plugin should be able to send EiffelSourceChangeCreatedEvent and EiffelSourceChangeSubmittedEvent
### Description

Dependent on #8 

The first draft of the gerrit plugin should be able to send EiffelSourceChangeCreatedEvent and EiffelSourceChangeSubmittedEvent. According to #6 there should be some kind of mechanism to generate the Eiffel events. The factory pattern could be one such way to realize this in order to make the one generating the event less affected by future additions/changes.

On initialization of the Eiffel event, it should populate itself with relevant fields from the gerrit event(besides from the links).

As a summary, this story should implement the functionality to generate Eiffel events as well as creating the actual EiffelSourceChangeCreatedEvent and EiffelSourceChangeSubmittedEvent object.

Please see #6 for the architectural overview document.

Christoffer-Cortes:
### Applicable Issues
closes #7 

### Description of the Change
- Add REMReM sender class
- Add remrem and rabbitmq container to docker-compose
- Fix start up of gerrit container with remote debugging

### Sign-off

    indicated in the file; or

    in the file; or

    it.

Signed-off-by: @Christoffer-Cortes @e-edling-ericsson


src/main/java/com/ericsson/gerrit/plugins/eiffel/messaging/EiffelEventSender.java
```diff
@@ -1,25 +1,121 @@
...
+        return isConfigurationSet;
+    }
+
+    private HttpRequest assembleRequest(HttpRequest httpRequest)
+            throws UnsupportedEncodingException {
+        final String username = pluginConfig.getRemremUsername();
+        final String password = pluginConfig.getRemremPassword();
+        final String url = pluginConfig.getRemremPublishURL();
+
+        //httpRequest.setBasicAuth(username, password);
```
m-linner-ericsson: Why leave this as a comment?

eiffel-gerrit-plugin-script
```diff
@@ -40,12 +42,14 @@ function do_rebuild {
   echo "######### BUILDING JARS ################"
   echo "########################################"
 
+  JAR_LOCATION=$(ls target/ | grep '^eiffel-gerrit-plugin-[0-9]*.[0-9]*.[0-9]*.jar')
   mvn clean install package -DskipTests
 
+  mv target/test-classes/env.bash target/env.bash
```
tobiasake: Why not copy these files directly from src/main/docker/... ?
Now Dockerfile is dependent on mvn goals that need to be executed first before build.
Maybe we want to be able to build based on released gerrit plugin jar files.

src/main/java/com/ericsson/gerrit/plugins/eiffel/messaging/EiffelEventSender.java
```diff
@@ -1,25 +1,121 @@
...
 
+    private void generateAndPublish()
+            throws URISyntaxException, IOException, EiffelEventSenderException {
+        if (!isConfigurationSet()) {
+            final String errorMessage = String.format(
+                    "Neccessary configuration is missing to send event."
+                            + "\neiffelMessage: %s\neiffelType: %s\neiffelProtocol: %s",
+                    eiffelMessage, eiffelType, EIFFEL_PROTOCOL);
+            throw new EiffelEventSenderException(errorMessage);
+        }
```
m-linner-ericsson: 1. You could break this out to its own method.
2. Why check this earlier?

src/main/java/com/ericsson/gerrit/plugins/eiffel/messaging/EiffelEventSender.java
```diff
@@ -1,25 +1,121 @@
...
+
+        if (HttpStatus.SC_OK == statusCode) {
+            LOGGER.info("Generated and published eiffel message successfully. \n{}", result);
+        } else {
+            final String errorMessage = String.format(
+                    "Could not generate and publish eiffel message due to server issue or invalid json data, "
+                            + "Status Code :: %d\npublishURL :: %s\ninput message :: %s\nError Message  :: %s",
+                    statusCode, httpRequest.getBaseUrl(), eiffelMessage, result);
+            throw new EiffelEventSenderException(errorMessage);
+        }
```
m-linner-ericsson: You could break this out to its own method

src/main/java/com/ericsson/gerrit/plugins/eiffel/messaging/EiffelEventSender.java
```diff
@@ -1,25 +1,121 @@
...
+    }
+
+    public EiffelEventSender setMessage(final EiffelEvent eiffelMessage) {
+        this.eiffelMessage = gson.toJsonTree(eiffelMessage).toString();
+        return this;
+    }
+
+    public EiffelEventSender setType(final String eiffelType) {
+        this.eiffelType = eiffelType;
+        return this;
```
m-linner-ericsson: If these parameters are mandatory. Why are they not part of the constructor?

src/main/java/com/ericsson/gerrit/plugins/eiffel/messaging/EiffelEventSender.java
```diff
@@ -1,25 +1,121 @@
...
     private final Gson gson = new Gson();
-    private final EiffelPluginConfiguration pluginConfig;
+    private String eiffelMessage;
+    private String eiffelType;
+    private EiffelPluginConfiguration pluginConfig;
+    private HttpRequest httpRequest;
 
     public EiffelEventSender(EiffelPluginConfiguration pluginConfig) {
+        this.httpRequest = new HttpRequest(HttpMethod.POST);
         this.pluginConfig = pluginConfig;
```
m-linner-ericsson: When having multiple constructors with variable number of parameters I thought it was standard to call the other with the missing parameter i.e.

```
EiffelEventSender(pluginConfig, httprequest)
```

eiffel-gerrit-plugin-script
Christoffer-Cortes: Good point but I think that is better left for another issue since it might require larger changes in the script file. We only changed what was needed to make the script work again since it wasn't propertly transferring the plugin file to the container and to set up the needed containers to test with remrem.

src/test/java/com/ericsson/gerrit/plugins/eiffel/messaging/EiffelEventSenderTest.java
```diff
@@ -0,0 +1,119 @@
...
+        } catch (Exception e) {
+            String exceptionType = e.getClass().getSimpleName();
+            String assertError = String.format(
+                    "No exception should have been thrown but was %s",
+                    exceptionType);
+            fail(assertError);
+        }
+    }
+
+    @Test
```
m-linner-ericsson: Is there a reason not to use the expected exception or rule tag (https://stackoverflow.com/questions/15216438/junit-testing-exceptions?noredirect=1&lq=1 - 1st or 2nd answer)

src/main/java/com/ericsson/gerrit/plugins/eiffel/messaging/EiffelEventSender.java
Christoffer-Cortes: Because the commented code is the true code, jitpack is troublesome and is not making eiffel-commons:1.0.1 available for download even though it claims it's ready... It's been hours now and I don't know what is going on.

src/main/java/com/ericsson/gerrit/plugins/eiffel/messaging/EiffelEventSender.java
Christoffer-Cortes: It was in it's own method before but was later moved in here. It was moved in here at the request of someone else. I don't care either way though it would perhaps be more proper to break it out and place it in before you call generateAndPublish method. @e-edling-ericsson could perhaps be convinced that breaking it out is better after all.

src/main/java/com/ericsson/gerrit/plugins/eiffel/messaging/EiffelEventSender.java
Christoffer-Cortes: So send HttpResponse to a verification method?

src/main/java/com/ericsson/gerrit/plugins/eiffel/messaging/EiffelEventSender.java
Christoffer-Cortes: The way I had originally designed it was to send them as a data model since I didn't want too many parameters but again this was changed at the request of others to the way it is now. I understand your point of view and I even programmed it in a similar fashion first but once again this change it at odds with other peoples requests. What am I to do? Hahaha....

src/main/java/com/ericsson/gerrit/plugins/eiffel/messaging/EiffelEventSender.java
Christoffer-Cortes: Yes that's true, we have done so before.

src/test/java/com/ericsson/gerrit/plugins/eiffel/messaging/EiffelEventSenderTest.java
Christoffer-Cortes: Yes I know about expected in the Test annotation and I've used at least on one occasion before. I can verify that the Exception happened for the right reason if I catch the exception myself.

Christoffer-Cortes: What I mean with right reason is that the exception one that can occur in two different scenarios. If I had a single type of exception for each scenario than we'd be sure without having to check further. Perhaps that is more desirable?

src/main/java/com/ericsson/gerrit/plugins/eiffel/messaging/EiffelEventSender.java
Christoffer-Cortes: I'll wait for dependency to be available before I merge.

src/main/java/com/ericsson/gerrit/plugins/eiffel/messaging/EiffelEventSender.java
Christoffer-Cortes: I broke it out and placed it before generateAndPublish

src/main/java/com/ericsson/gerrit/plugins/eiffel/messaging/EiffelEventSender.java
Christoffer-Cortes: This is now placed in verifyResponse method.

src/main/java/com/ericsson/gerrit/plugins/eiffel/messaging/EiffelEventSender.java
Christoffer-Cortes: This one I don't know what to do about, I think we will just leave it for now.

src/main/java/com/ericsson/gerrit/plugins/eiffel/messaging/EiffelEventSender.java
Christoffer-Cortes: The first constructor now calls the second one.

src/test/java/com/ericsson/gerrit/plugins/eiffel/messaging/EiffelEventSenderTest.java
Christoffer-Cortes: This one I'm not sure about, perhaps I'll make two separate exceptions so there doesn't need to be more than an expected exception.

Christoffer-Cortes: I added separate exceptions and removed all asserts and let the exception throws decide if a test has succeeded or not.

eiffel-gerrit-plugin-script
e-edling-ericsson: I suppose that docker is running from the target folder. Its not possible to go "../" in that context. That would be my guess why it isn't taking it from the src/main/docker.

src/main/java/com/ericsson/gerrit/plugins/eiffel/messaging/EiffelEventSender.java
e-edling-ericsson: Yes I think it's better to break it out to "verifyConfiguration" and put it right before "generateAndPublish"

eiffel-gerrit-plugin-script
tobiasake: We run Docker build from git root in other project as context path, why not do the same here.

Christoffer-Cortes: I still suggest to make an issue about improving the test/start script since it's not directly relevant to this PR.

Christoffer-Cortes: I made an issue https://github.com/eiffel-community/eiffel-gerrit-plugin/issues/30
I'm not sure about the details since it's not my suggestion... I just created it.

tobiasake: Ok

src/main/docker/env.bash
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
+
+if [[ "${DOCKER_HOST}" ]]; then
+  export HOST=$(echo ${DOCKER_HOST} | sed -e 's|^tcp\://||;s|:.*$||');
```
tobiasake: What does DOCKER_HOST varaible from? It looks like regex is extracting host name from a tcp address.
Anyone that knows what and when DOCKER_HOST is set and used.

src/main/java/com/ericsson/gerrit/plugins/eiffel/listeners/PatchsetCreatedEventListener.java
```diff
@@ -64,8 +64,9 @@ protected void prepareAndSendEiffelEvent(Event gerritEvent,
         EiffelSourceChangeCreatedEvent eiffelEvent = EiffelSourceChangeCreatedEventGenerator.generate(
                 patchSetCreatedEvent, pluginConfig);
         EiffelEventSender eiffelEventSender = new EiffelEventSender(pluginConfig);
-        eiffelEventSender.send(eiffelEvent);
-
+        eiffelEventSender.setType("EiffelSourceChangeCreatedEvent");
```
a-breid-ericsson: Please remove the hardcoded type here. Type is the same as the Event Class. One could use EventClass....class.getName() or something.

Or move the type to a static variable.

src/main/java/com/ericsson/gerrit/plugins/eiffel/messaging/EiffelEventSender.java
```diff
@@ -1,25 +1,140 @@
...
+            LOGGER.error("Failed to send eiffel message.", e);
+        }
+    }
+
+    public EiffelEventSender setMessage(final EiffelEvent eiffelMessage) {
+        this.eiffelMessage = gson.toJsonTree(eiffelMessage).toString();
+        return this;
+    }
+
+    public EiffelEventSender setType(final String eiffelType) {
```
a-breid-ericsson: Some Javadoc here.

src/main/java/com/ericsson/gerrit/plugins/eiffel/messaging/EiffelEventSender.java
```diff
@@ -1,25 +1,140 @@
...
+
+    private void generateAndPublish()
+            throws URISyntaxException, IOException, MissingConfigurationException,
+            HttpRequestFailedException {
+        assembleRequest();
+        final ResponseEntity response = httpRequest.performRequest();
+        verifyResponse(response);
     }
 
+    private void verifyResponse(ResponseEntity response) throws HttpRequestFailedException {
```
a-breid-ericsson: Do we want to have retries?

pom.xml
```diff
@@ -45,6 +54,24 @@
             <scope>test</scope>
         </dependency>
 
+        <dependency>
+            <groupId>org.springframework</groupId>
```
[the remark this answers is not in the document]
Christoffer-Cortes: I think it was needed for Autowired annotation.

src/main/docker/env.bash
Christoffer-Cortes: I just took the env.bash from EI. I can remove the if statement.

src/main/java/com/ericsson/gerrit/plugins/eiffel/listeners/PatchsetCreatedEventListener.java
Christoffer-Cortes: Ah true, I think it's getSimpleName otherwise you get the package name as well, but yes we could do that.

src/main/java/com/ericsson/gerrit/plugins/eiffel/messaging/EiffelEventSender.java
Christoffer-Cortes: Well I can only speak for myself but I think that REMReM should be responsible to retries. I don't think it currently does this but it should be part of that service.

src/main/docker/env.bash
tobiasake: I can't find that in EI env.bash file:
https://github.com/eiffel-community/eiffel-intelligence/blob/master/src/main/docker/env.bash
If its not used or we don't know what its intended to be used for, then remove that line the if state and keep only code that is in else case.

src/main/java/com/ericsson/gerrit/plugins/eiffel/messaging/EiffelEventSender.java
a-breid-ericsson: It is not clear to me what Type is.

src/main/java/com/ericsson/gerrit/plugins/eiffel/messaging/EiffelEventSender.java
a-breid-ericsson: Will not work if REMReM is down :)

pom.xml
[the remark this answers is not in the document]
a-breid-ericsson: Ther is no @Autowired  or any @Component to @Autowire in the code.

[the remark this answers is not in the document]
Christoffer-Cortes: You are right, this was only used in the jira plugin. I will remove it.

src/main/docker/env.bash
Christoffer-Cortes: Done

src/main/java/com/ericsson/gerrit/plugins/eiffel/messaging/EiffelEventSender.java
Christoffer-Cortes: So "Set the eiffel event type", that is the sort of sentence that I think doesn't add anything useful. Do you have anything in mind for the javadoc? Perhaps the method could be called something different if you feel setType is not enough?

src/main/java/com/ericsson/gerrit/plugins/eiffel/messaging/EiffelEventSender.java
Christoffer-Cortes: I see, I assumed you were talking about sending the message out to the mb. There is something called HttpRequestRetryHandler in apache http lib that might be useful for failed requests. This would require changes in eiffel-commons though. My only apprehension is that I'm not sure what the benefit of using remrem is then...

src/main/java/com/ericsson/gerrit/plugins/eiffel/messaging/EiffelEventSender.java
e-edling-ericsson: That should have added a lot, but then i think the name is not good enough. Maybe the name should be, setEiffelEventType then?

src/main/java/com/ericsson/gerrit/plugins/eiffel/messaging/EiffelEventSender.java
e-edling-ericsson: Is remrem down, we won't be able to do much. Although, we can't know if its simply down or some other temporary network issue occured, i think we should have some kind of retry

src/main/java/com/ericsson/gerrit/plugins/eiffel/messaging/EiffelEventSender.java
Christoffer-Cortes: I named them more specifically now to avoid using javadoc.

src/main/java/com/ericsson/gerrit/plugins/eiffel/messaging/EiffelEventSender.java
Christoffer-Cortes: This implementation should a configurable option in HttpRequest then. I can create an issue on this so that the scope does not grow on this one.

Christoffer-Cortes: Issue https://github.com/eiffel-community/eiffel-commons/issues/29

eiffel-gerrit-plugin-script
Christoffer-Cortes: I think this is fixed.

Issue #31, opened 2019-10-21
Add retry handling when sending Eiffel message to REMReM


Christoffer-Cortes:
### Description
We need to implement some sort of thread safe retry handling for our requests.

### Motivation
We don't want to immediately dismiss messages that don't reach REMReM.

### Exemplification
[Spring-retry](https://www.baeldung.com/spring-retry)
Spring has a solution for a retry mechanism.
[CircuitBreaker pattern](https://en.wikipedia.org/wiki/Circuit_breaker_design_pattern)
[CircuitBreaker Spring](https://spring.io/guides/gs/circuit-breaker/)
Also look at circuit breaker pattern and the implementation by spring.

### Benefits
Avoid lost requests.

### Possible Drawbacks
It could potentially put an extra burden on resources and make the system more unresponsive.
This needs to be implemented carefully.

PR #32, opened 2019-10-21
Add retry support for REMReM requests

Christoffer-Cortes:
### Applicable Issues
Closes https://github.com/eiffel-community/eiffel-gerrit-plugin/issues/31

### Description of the Change
This solution uses Spring-retry to send multiple request attempts to REMReM.
A retry configuration class contains the RetryTemplate bean with the settings for the retry mechanism.
This RetryTemplate is then used to wrap the send method and catches the RuntimeExceptions thrown by the sender when a http request fails.
The RuntimeException is what triggers the retry.

TODO: circuit breaker pattern needs to be investigated and other questions that might arise.

### Sign-off

    indicated in the file; or

    in the file; or

    it.

Signed-off-by: @Christoffer-Cortes

### DIFF
--- a/src/main/java/com/ericsson/gerrit/plugins/eiffel/listeners/ChangeMergedEventListener.java
+++ b/src/main/java/com/ericsson/gerrit/plugins/eiffel/listeners/ChangeMergedEventListener.java
@@ -29,6 +29,7 @@ import org.springframework.retry.support.RetryTemplate;
 import com.ericsson.gerrit.plugins.eiffel.configuration.EiffelPluginConfiguration;
 import com.ericsson.gerrit.plugins.eiffel.events.EiffelSourceChangeSubmittedEvent;
 import com.ericsson.gerrit.plugins.eiffel.events.generators.EiffelSourceChangeSubmittedEventGenerator;
+import com.ericsson.gerrit.plugins.eiffel.exceptions.HttpRequestFailedException;
 import com.ericsson.gerrit.plugins.eiffel.messaging.EiffelEventSender;
 import com.google.gerrit.extensions.annotations.PluginData;
 import com.google.gerrit.extensions.annotations.PluginName;
@@ -71,7 +72,7 @@ public class ChangeMergedEventListener extends AbstractEventListener {
         EiffelEventSender eiffelEventSender = new EiffelEventSender(pluginConfig);
         eiffelEventSender.setEiffelEventType(eiffelEvent.getClass().getSimpleName());
         eiffelEventSender.setEiffelEventMessage(eiffelEvent);
-        retryTemplate.execute(new RetryCallback<Void, RuntimeException>() {
+        retryTemplate.execute(new RetryCallback<Void, HttpRequestFailedException>() {
             @Override
             public Void doWithRetry(RetryContext context) {
                 eiffelEventSender.send();
--- a/src/main/java/com/ericsson/gerrit/plugins/eiffel/listeners/ChangeMergedEventListener.java
+++ b/src/main/java/com/ericsson/gerrit/plugins/eiffel/listeners/ChangeMergedEventListener.java
@@ -21,6 +21,10 @@ import java.io.File;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.retry.RetryCallback;
+import org.springframework.retry.RetryContext;
+import org.springframework.retry.support.RetryTemplate;
 
 import com.ericsson.gerrit.plugins.eiffel.configuration.EiffelPluginConfiguration;
 import com.ericsson.gerrit.plugins.eiffel.events.EiffelSourceChangeSubmittedEvent;
@@ -40,6 +44,9 @@ import com.google.inject.Inject;
 public class ChangeMergedEventListener extends AbstractEventListener {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(ChangeMergedEventListener.class);
+
+    @Autowired
+    private RetryTemplate retryTemplate;
 
     @Inject
     public ChangeMergedEventListener(@PluginName final String pluginName,
@@ -64,6 +71,12 @@ public class ChangeMergedEventListener extends AbstractEventListener {
         EiffelEventSender eiffelEventSender = new EiffelEventSender(pluginConfig);
         eiffelEventSender.setEiffelEventType(eiffelEvent.getClass().getSimpleName());
         eiffelEventSender.setEiffelEventMessage(eiffelEvent);
-        eiffelEventSender.send();
+        retryTemplate.execute(new RetryCallback<Void, RuntimeException>() {
+            @Override
+            public Void doWithRetry(RetryContext context) {
+                eiffelEventSender.send();
+                return null;
+            }
+        });
     }
 }
--- a/src/main/java/com/ericsson/gerrit/plugins/eiffel/listeners/ChangeMergedEventListener.java
+++ b/src/main/java/com/ericsson/gerrit/plugins/eiffel/listeners/ChangeMergedEventListener.java
@@ -21,16 +21,10 @@ import java.io.File;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.retry.RetryCallback;
-import org.springframework.retry.RetryContext;
-import org.springframework.retry.support.RetryTemplate;
 
 import com.ericsson.gerrit.plugins.eiffel.configuration.EiffelPluginConfiguration;
 import com.ericsson.gerrit.plugins.eiffel.events.EiffelSourceChangeSubmittedEvent;
 import com.ericsson.gerrit.plugins.eiffel.events.generators.EiffelSourceChangeSubmittedEventGenerator;
-import com.ericsson.gerrit.plugins.eiffel.exceptions.HttpRequestFailedException;
-import com.ericsson.gerrit.plugins.eiffel.messaging.EiffelEventSender;
 import com.google.gerrit.extensions.annotations.PluginData;
 import com.google.gerrit.extensions.annotations.PluginName;
 import com.google.gerrit.server.events.ChangeMergedEvent;
@@ -45,9 +39,6 @@ import com.google.inject.Inject;
 public class ChangeMergedEventListener extends AbstractEventListener {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(ChangeMergedEventListener.class);
-
-    @Autowired
-    private RetryTemplate retryTemplate;
 
     @Inject
     public ChangeMergedEventListener(@PluginName final String pluginName,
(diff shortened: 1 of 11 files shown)


magnusbaeck:
Doesn't Gerrit run the event listeners synchronously, i.e. blocking an event listener with a retry will block whatever Gerrit was doing that triggered the event (like a push operation). I'd expect that to be undesirable for many of not most users. Wouldn't it make more sense to have a limited in-memory queue like the [rabbitmq plugin](https://gerrit.googlesource.com/plugins/rabbitmq/)?

Christoffer-Cortes:
> Doesn't Gerrit run the event listeners synchronously, i.e. blocking an event listener with a retry will block whatever Gerrit was doing that triggered the event (like a push operation). I'd expect that to be undesirable for many of not most users. Wouldn't it make more sense to have a limited in-memory queue like the [rabbitmq plugin](https://gerrit.googlesource.com/plugins/rabbitmq/)?

I assumed that the listener was a threaded operation but perhaps it is not.
We have some uncertainties around retries etc. even in REMReM. I'm not sure how it is handled in that service.
I will take a look at this anyhow and see how we can make this safer and without blocking events.

magnusbaeck:
> I assumed that the listener was a threaded operation but perhaps it is not.

One of my colleagues suggested that event dispatching is non-blocking as of Gerrit 2.16 or 3.0 (a change Martin Fick supposedly was involved in), but it seems you want to support earlier releases as well. I haven't dug into the commit logs or release notes to verify that claim.

src/main/java/com/ericsson/gerrit/plugins/eiffel/listeners/ChangeMergedEventListener.java
```diff
@@ -64,6 +72,12 @@ protected void prepareAndSendEiffelEvent(Event gerritEvent,
         EiffelEventSender eiffelEventSender = new EiffelEventSender(pluginConfig);
         eiffelEventSender.setEiffelEventType(eiffelEvent.getClass().getSimpleName());
         eiffelEventSender.setEiffelEventMessage(eiffelEvent);
-        eiffelEventSender.send();
+        retryTemplate.execute(new RetryCallback<Void, HttpRequestFailedException>() {
+            @Override
+            public Void doWithRetry(RetryContext context) {
+                eiffelEventSender.send();
+                return null;
```
m-linner-ericsson: Do we need to return null?

src/main/java/com/ericsson/gerrit/plugins/eiffel/listeners/ChangeMergedEventListener.java
```diff
@@ -64,6 +72,12 @@ protected void prepareAndSendEiffelEvent(Event gerritEvent,
...
+            }
+        });
```
m-linner-ericsson: This code is duplicated in two classes. Can't we move it to the event sender or the base class

src/main/java/com/ericsson/gerrit/plugins/eiffel/listeners/PatchsetCreatedEventListener.java
```diff
@@ -66,7 +74,13 @@ protected void prepareAndSendEiffelEvent(Event gerritEvent,
...
         eiffelEventSender.setEiffelEventType(eiffelEvent.getClass().getSimpleName());
         eiffelEventSender.setEiffelEventMessage(eiffelEvent);
-        eiffelEventSender.send();
+        retryTemplate.execute(new RetryCallback<Void, HttpRequestFailedException>() {
+            @Override
+            public Void doWithRetry(RetryContext context) {
+                eiffelEventSender.send();
+                return null;
+            }
+        });
```
m-linner-ericsson: Code duplication see comment above

src/test/java/com/ericsson/gerrit/plugins/eiffel/messaging/RetryRequestTest.java
```diff
@@ -0,0 +1,125 @@
...
+        EiffelEventSender sender = new EiffelEventSender(pluginConfig, httpRequest);
+        sender.setEiffelEventMessage(new EiffelSourceChangeCreatedEvent());
+        sender.setEiffelEventType(EIFFEL_TYPE);
+        retryTemplate.execute(new RetryCallback<Void, HttpRequestFailedException>() {
+            @Override
+            public Void doWithRetry(RetryContext context) {
+                counter.setValue(counter.getValue().intValue() + 1);
+                sender.send();
+                return null;
+            }
```
m-linner-ericsson: Good that you test the logic but you are not testing the cod in `PatchsetCreatedEventListener` and `ChangeMergedEventListener`

src/main/java/com/ericsson/gerrit/plugins/eiffel/messaging/EiffelEventSender.java
```diff
@@ -60,16 +60,19 @@ public EiffelEventSender(EiffelPluginConfiguration pluginConfig, HttpRequest htt
     }
 
     /**
-     * Sends a REMReM eiffel message using
+     * Sends a REMReM Eiffel message to the generateAndPublish endpoint. RuntimeException is thrown
+     * when a HttpRequestFailedException occurs so that the retry logic works.
```
m-linner-ericsson: I don't get the comment. I guess `HttpRequestFailedException` triggers the retry logic so that the reason why we need to throw it?

src/main/java/com/ericsson/gerrit/plugins/eiffel/messaging/EiffelEventSender.java
```diff
@@ -60,16 +60,19 @@ public EiffelEventSender(EiffelPluginConfiguration pluginConfig, HttpRequest htt
...
             verifyConfiguration();
             generateAndPublish();
-        } catch (URISyntaxException | IOException | MissingConfigurationException
-                | HttpRequestFailedException e) {
-            LOGGER.error("Failed to send eiffel message.", e);
+        } catch (URISyntaxException | IOException | MissingConfigurationException e1) {
+            LOGGER.error("Failed to send eiffel message.", e1);
+        } catch (HttpRequestFailedException e2) {
+            LOGGER.error("Failed to send eiffel message.", e2);
+            throw e2;
```
magnusbaeck: I think there are two flaws with this logic:

* We're unconditionally retrying all non-200 HTTP codes, including various 4xx errors that aren't likely to get fixed with us retrying (e.g. 400, 405, and 409).
* IOException errors should in many (most?) cases be retried since they could very well be the result of ETIMEDOUT, ECONNREFUSED, or EHOSTUNREACH errors, or temporary DNS lookup problems.

sselberg:
> > I assumed that the listener was a threaded operation but perhaps it is not.
> 
> One of my colleagues suggested that event dispatching is non-blocking as of Gerrit 2.16 or 3.0 (a change Martin Fick supposedly was involved in), but it seems you want to support earlier releases as well. I haven't dug into the commit logs or release notes to verify that claim.

It's definitely blocking  the event-bus in 2.14 and 2.15. That's why I implemented the listener -> queue -> publisher pattern in the rabbitmq plugin, I suggest you take a similar approach that does not depend on external resources to be able to handle the event.
From a quick search I cannot find any changes in Gerrit core that matches these criteria. Could it be changes in the events plugin that was referenced?

src/test/java/com/ericsson/gerrit/plugins/eiffel/messaging/RetryRequestTest.java
Christoffer-Cortes: testPrepareAndSendEiffelEvent tests the ChangeMergedEventListener. I don't do the same for PatchsetCreatedEventListener though.

src/main/java/com/ericsson/gerrit/plugins/eiffel/listeners/ChangeMergedEventListener.java
Christoffer-Cortes: This is a requirement of the doWithRetry method. I wont accept an implementation without an a null return. Has something to do with Void probably.

src/main/java/com/ericsson/gerrit/plugins/eiffel/messaging/EiffelEventSender.java
Christoffer-Cortes: Ah yes, this has been changed, HttpRequestFailedException is of the type RuntimeException now. I must update the docs.

src/main/java/com/ericsson/gerrit/plugins/eiffel/messaging/EiffelEventSender.java
Christoffer-Cortes: Yes, previously when I played around the apache retry logic I had more code checks, but this was just implemented on top of the current code without further changes to the Exception handling. There is also 401 where you might not want to retry if authorization fails etc. I will look more into this and consult my colleagues.

magnusbaeck: Indeed, 401 shouldn't be retried (hello, instant account lockout) but one could make the argument that 403 errors _should_ be retried to better cope with temporary misconfigurations.

src/test/java/com/ericsson/gerrit/plugins/eiffel/messaging/RetryRequestTest.java
m-linner-ericsson: If you move the duplicated code you could only have one place to test

Christoffer-Cortes:
> > > I assumed that the listener was a threaded operation but perhaps it is not.
> > 
> > 
> > One of my colleagues suggested that event dispatching is non-blocking as of Gerrit 2.16 or 3.0 (a change Martin Fick supposedly was involved in), but it seems you want to support earlier releases as well. I haven't dug into the commit logs or release notes to verify that claim.
> 
> It's definitely blocking the event-bus in 2.14 and 2.15. That's why I implemented the listener -> queue -> publisher pattern in the rabbitmq plugin, I suggest you take a similar approach that does not depend on external resources to be able to handle the event.
> From a quick search I cannot find any changes in Gerrit core that matches these criteria. Could it be changes in the events plugin that was referenced?

@erik-edling and I checked the changeMergedListener and started to submits on gerrit with a breakpoint inside prepareAndSendEiffelEvent. Gerrit spawned a thread for each event triggered resulting in two separate threads, so it does not appear to block subsequent events. Are you sure this is the case and if so, how could we reproduce this?

sselberg:
> @erik-edling and I checked the changeMergedListener and started two submits on gerrit with a breakpoint inside prepareAndSendEiffelEvent. Gerrit spawned a thread for each event triggered resulting in two separate threads, so it does not appear to block subsequent events. Are you sure this is the case and if so, how could we reproduce this?

So after this you should have two threads that are blocked on prepareAndSendEiffelEvent.
What happens if you keep the breakpoint and do 1000 additional submits?

src/main/java/com/ericsson/gerrit/plugins/eiffel/listeners/ChangeMergedEventListener.java
Christoffer-Cortes: It's now called from a shared method in the parent.

src/main/java/com/ericsson/gerrit/plugins/eiffel/listeners/PatchsetCreatedEventListener.java
Christoffer-Cortes: Done.

src/test/java/com/ericsson/gerrit/plugins/eiffel/messaging/RetryRequestTest.java
Christoffer-Cortes: I test the send method in AbstractEventListener now.

Christoffer-Cortes:
@sselberg @magnusbaeck I've written an issue [here](https://github.com/eiffel-community/eiffel-gerrit-plugin/issues/36). I've not had time to look into the rabbitmq plugin yet but I imagine we need to queue up jobs that are worked by a thread executor or something like that. For now we'll review and merge this as is and I will continue the work on the other issue. Thanks for bringing this to our attention.

Issue #12, opened 2019-09-17
Implement state and DB handling mechanism


erik-edling:
### Description

Dependent on #8 

The eiffel events should be able to ask the state handling mechanism about the eiffel id for a specific link type by providing some input(to be decided in this story) in order for the state handler to know the current context for the link(branch etc).

This story also includes implementation of the DB handling mechanism where all the eiffel event IDs are stored. It should be possible to save and retrieve eiffel event ids. 

Please see #6 for the architectural overview document.

m-linner-ericsson:
Would this text add more information to the issue or is it too detailed?

Input:
- Gerrit event
- Eiffel event
- Storage location

Outputs:
- Previous event id

Public interface:
- Get previous id
- Store Eiffel event

Error handling:
- Store fails
- Read fails

Possible usage:
```
StateHandler stateHandler = new StateHandler(storageLocation);

// Get previous id
stateHandler.getPreviousEiffelEventId(gerritEvent);

// Store events
stateHandler.storeEiffelEvent(eiffelEvent);
```

a-breid-ericsson:
In this case it may be too detailed since the person investigating the issue maybe should present input, output and interfaces.

m-linner-ericsson:
Just noticing that this will limit the amount of people being able to help out...

PR #37, opened 2019-10-29
Add Linking and database handling

erik-edling:
### Applicable Issues
#12

### Description of the Change
- Add database handling
- Add linking

### Sign-off

    indicated in the file; or

    in the file; or

    it.

Signed-off-by:  @saif-ericsson  @erik-edling


src/main/java/com/ericsson/gerrit/plugins/eiffel/state/StateFactory.java
```diff
@@ -0,0 +1,21 @@
...
+            case "EiffelSourceChangeCreatedEvent":
+                return new SourceChangeCreatedState(pluginDir);
+            case "EiffelSourceChangeSubmittedEvent":
+                return new SourceChangeSubmittedState(pluginDir);
+            }
+        } catch (Exception e) {
+            // TODO Change this
+            return null;
+        }
+        return null;
```
erik-edling: We will change this part, i suppose this is the fate when creating TODOs :)

src/main/java/com/ericsson/gerrit/plugins/eiffel/handlers/StateHandler.java
```diff
@@ -0,0 +1,187 @@
...
+     * @param project
+     * @param branch
+     * @return
+     * @throws SomeRunTimeException
+     */
+    public String getLastSourceChangeSubmittedEiffelEvent(String project, String branch) throws NoSuchElementException {
+        return getLastCreatedEiffelEvent(project, branch, Table.SCS_TABLE);
+    }
+
+    public void setState(String project, String branch, String changeId, EiffelEvent eiffelEvent) {
```
a-breid-ericsson: Do we need to save the whole Eiffel event? the DB files would grow in size quickly. all we need to save is the EventIds

src/main/java/com/ericsson/gerrit/plugins/eiffel/state/SourceChangeCreatedState.java
```diff
@@ -0,0 +1,33 @@
...
+import java.sql.SQLException;
+
+import com.ericsson.gerrit.plugins.eiffel.events.EiffelEvent;
+import com.ericsson.gerrit.plugins.eiffel.events.EiffelSourceChangeCreatedEvent;
+import com.ericsson.gerrit.plugins.eiffel.handlers.NoSuchElementException;
+import com.ericsson.gerrit.plugins.eiffel.handlers.Table;
+
+public class SourceChangeCreatedState extends State {
+
+    public SourceChangeCreatedState(File pluginDir) {
```
a-breid-ericsson: This PR adds States for each event type, and StateAccesors for each event type, if there will ever be more than 2 event types this will become many many files. It also seems like there is duplicated code on most of the files. This could be done better, maybe there should be a team discussion about this.

src/main/java/com/ericsson/gerrit/plugins/eiffel/state/State.java
```diff
@@ -0,0 +1,131 @@
...
+        }
+    }
+
+    /**
+     * Creates parent directories of a project if they don't exist and is included
+     * in the project name.
+     *
+     * @param path
+     */
+    protected void createParentDirsIfNecessary(String path) {
```
a-breid-ericsson: State should maybe not handle creation of path, this should be closer to the DB handler

src/main/java/state/SourceChangeCreatedStateAccessor.java
```diff
@@ -0,0 +1,26 @@
...
+        super(pluginDir, project, branch, changeId);
+    }
+
+    @Override
+    public String getEventId() throws NoSuchElementException {
+        return getLastSubmittedEiffelEvent(pluginDir, project, branch, Table.SCS_TABLE);
+    }
+
+    @Override
+    public void setState(String eiffelEventId)
```
a-breid-ericsson: You set EventId not a state.

src/main/java/state/StateAccessor.java
```diff
@@ -0,0 +1,132 @@
...
+        this.project = project;
+        this.branch = branch;
+    }
+
+    public abstract String getEventId() throws NoSuchElementException;
+
+    public abstract void setState(String eiffelEventId)
+            throws NoSuchElementException, ConnectException, SQLException;
+
+    protected String getLastSubmittedEiffelEvent(File pluginDir, String project, String tableColumnName,
```
a-breid-ericsson: Is this code not duplicated from StateHandler

src/main/java/state/StateFactory.java
```diff
@@ -0,0 +1,34 @@
...
+            case "EiffelSourceChangeCreatedEvent":
+                EiffelSourceChangeCreatedEvent eiffelSourceChangeCreatedEvent = (EiffelSourceChangeCreatedEvent) eiffelEvent;
+                String project = eiffelSourceChangeCreatedEvent.eventParams.data.gitIdentifier.repoName;
+                String branch = eiffelSourceChangeCreatedEvent.eventParams.data.gitIdentifier.branch;
+                String changeId = eiffelSourceChangeCreatedEvent.eventParams.data.change.id;
+
+                return new SourceChangeCreatedStateAccessor(pluginDir, project, branch, changeId);
+            case "EiffelSourceChangeSubmittedEvent":
+                EiffelSourceChangeSubmittedEvent eiffelSourceChangeSubmittedEvent = (EiffelSourceChangeSubmittedEvent) eiffelEvent;
+                String project1 = eiffelSourceChangeSubmittedEvent.eventParams.data.gitIdentifier.repoName;
```
a-breid-ericsson: Please dont have variables named xxx1 xxx2 and so on.

src/main/java/state/StateFactory.java
```diff
@@ -0,0 +1,34 @@
+package state;
+
+import java.io.File;
+
+import com.ericsson.gerrit.plugins.eiffel.events.EiffelEvent;
+import com.ericsson.gerrit.plugins.eiffel.events.EiffelSourceChangeCreatedEvent;
+import com.ericsson.gerrit.plugins.eiffel.events.EiffelSourceChangeSubmittedEvent;
+
+public class StateFactory {
+    public static StateAccessor getStateAccessor(File pluginDir, EiffelEvent eiffelEvent) {
```
a-breid-ericsson: In State yu have set event id, and get event id, I dont see the point of adding more layers just for Accessing a state when you already have the state and you should be able to get id from that. Removing Accessors would reduce the need of adding 4 classes for each event type to 3.

src/test/java/com/ericsson/gerrit/plugins/eiffel/events/generators/EiffelEventGeneratorTest.java
```diff
@@ -71,22 +71,24 @@ public void testEiffelSourceChangeSubmittedEventGenerator() {
         setUpMocks();
         populateChangeMergedEvent();
 
-        EiffelSourceChangeSubmittedEvent eiffelEvent = EiffelSourceChangeSubmittedEventGenerator.generate(
-                changeMergedEvent, pluginConfig);
+        // EiffelSourceChangeSubmittedEvent eiffelEvent =
```
a-breid-ericsson: Please activate tests or remove commented code.

src/test/java/com/ericsson/gerrit/plugins/eiffel/handlers/StateHandlerTest.java
```diff
@@ -0,0 +1,100 @@
...
+
+import org.junit.runner.RunWith;
+import org.powermock.core.classloader.annotations.PrepareForTest;
+import org.powermock.modules.junit4.PowerMockRunner;
+
+@RunWith(PowerMockRunner.class)
+@PrepareForTest(fullyQualifiedNames = "com.ericsson.gerrit.plugins.eiffel.*")
+public class StateHandlerTest {
+
+    /*private static final String FILE_ENDING = "db";
```
a-breid-ericsson: Please activate tests or remove commented code.

src/test/java/com/ericsson/gerrit/plugins/eiffel/messaging/EiffelEventSenderTest.java
```diff
@@ -33,36 +26,37 @@ public void beforeTest() throws IOException, URISyntaxException {
...
-    public void testEventSenderWithBadStatus() throws Exception {
-        setUpMockActionsWithBadStatus();
-
-        EiffelEventSender sender = new EiffelEventSender(pluginConfig, httpRequest);
-        sender.setEiffelEventMessage(new EiffelSourceChangeCreatedEvent());
-        sender.setEiffelEventType(EIFFEL_TYPE);
-
-        Whitebox.invokeMethod(sender, "generateAndPublish");
-    }
+    // @Test
```
a-breid-ericsson: Please activate tests or remove commented code.

src/main/java/com/ericsson/gerrit/plugins/eiffel/events/generators/EiffelEventGenerator.java
```diff
@@ -51,6 +60,32 @@ protected static String createRepoURI(String url, String projectName) {
         }
     }
 
+    protected static String getLastSourceChangeSubmittedEiffelEventId(String projectName, String branch,
```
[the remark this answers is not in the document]
erik-edling: I've already cleaned this one up and will be available in the next push :)

src/main/java/com/ericsson/gerrit/plugins/eiffel/events/generators/EiffelSourceChangeCreatedEventGenerator.java
```diff
@@ -49,6 +60,9 @@ public static EiffelSourceChangeCreatedEvent generate(PatchSetCreatedEvent patch
         final int insertions = patchSetAttribute.sizeInsertions;
         final int deletions = patchSetAttribute.sizeDeletions;
         final String changeId = patchSetCreatedEvent.changeKey.toString();
+        final Link previousVersionLink = createPreviousVersionLink(pluginDirectoryPath, projectName,
```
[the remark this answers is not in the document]
erik-edling: That would apply with the rest of the data as well but it would become to messy in my opinion. That's why we have one section where we retrieve all data and one section where we populate it.

src/main/java/com/ericsson/gerrit/plugins/eiffel/events/generators/EiffelSourceChangeCreatedEventGenerator.java
```diff
@@ -71,15 +85,44 @@ public static EiffelSourceChangeCreatedEvent generate(PatchSetCreatedEvent patch
...
 
-        // TODO
-        // String latestEiffelSourceChangeCreatedEventId =
-        // getLatestEiffelSourceChangeCreatedEventId();
-        // setPreviousVersionLink(latestEiffelSourceChangeCreatedEventId);
-
-        // String latestEiffelSourceChangeSubmittedEventId =
-        // getLatestEiffelSourceChangeSubmittedEventId();
-        // setBaseLink(latestEiffelSourceChangeSubmittedEventId);
+        if (previousVersionLink != null) {
```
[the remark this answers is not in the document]
erik-edling: We only want this to apply when its null. If the value is "". Then it should have that value in the links.

src/main/java/com/ericsson/gerrit/plugins/eiffel/events/generators/EiffelSourceChangeCreatedEventGenerator.java
```diff
@@ -71,15 +85,44 @@ public static EiffelSourceChangeCreatedEvent generate(PatchSetCreatedEvent patch
...
+            Link baseLink = new Link();
+            baseLink.type = LINK_TYPE_BASE;
+            baseLink.target = lastSourceChangeSubmitted;
+
+            return baseLink;
+        }
+        return null;
+    }
+
+    private static Link createPreviousVersionLink(File pluginDirectoryPath, final String projectName,
```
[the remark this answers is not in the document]
erik-edling: I've already cleaned this one up and will be available in the next push :)

src/main/java/state/StateFactory.java
erik-edling: Its removed in the new commit because of some restructuring.

src/test/java/com/ericsson/gerrit/plugins/eiffel/events/generators/EiffelEventGeneratorTest.java
erik-edling: We wrote in the PR that we will add the tests so skip them for now, we will send update in this PR for the tests.

src/main/java/state/StateFactory.java
erik-edling: I have no idea why the accessor still exists. we renamed it. It should be completely removed :)

pom.xml
```diff
@@ -53,6 +53,11 @@
             <version>1.7.26</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
```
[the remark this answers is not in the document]
erik-edling: done in upcoming change

src/main/java/com/ericsson/gerrit/plugins/eiffel/events/generators/EiffelEventGenerator.java
```diff
@@ -51,6 +63,33 @@ protected static String createRepoURI(String url, String projectName) {
         }
     }
 
+    protected static Link createLink(String linkType, String linkedEiffelEventType, File pluginDirectoryPath, final String projectName,
```
SantoshNC68: the parameters can be final

src/main/java/com/ericsson/gerrit/plugins/eiffel/events/generators/EiffelEventGenerator.java
```diff
@@ -51,6 +63,33 @@ protected static String createRepoURI(String url, String projectName) {
...
+            Link link = new Link();
+            link.type = linkType;
+            link.target = lastEiffelEvent;
+
+            return link;
+        }
+        return null;
+    }
+
+    protected static String getEiffelEventIdFromState(State state, String projectName, String tableColumnName) {
```
SantoshNC68: same as above

src/main/java/com/ericsson/gerrit/plugins/eiffel/handlers/DataBaseHandler.java
```diff
@@ -0,0 +1,245 @@
...
+ * function.
+ *
+ */
+public class DataBaseHandler {
+    private static final Logger LOGGER = LoggerFactory.getLogger(DataBaseHandler.class);
+    private static final String EVENT_ID_KEY = "eventId";
+    private final String databaseFile;
+
+    /**
+     * Constructor that takes plugin directory path as String and plugin name as
```
SantoshNC68: The javadoc seems a bit confusing as it is saying Plugin name where as the parameter says filename.
filename, N can be capitalized.

either use fileName or pluginName.

src/main/java/com/ericsson/gerrit/plugins/eiffel/handlers/DataBaseHandler.java
```diff
@@ -0,0 +1,245 @@
...
+
+    /**
+     * Constructor that takes plugin directory path as File and plugin name as
+     * String.
+     *
+     * @param pluginDir
+     * @param filename
+     * @throws ConnectException
+     */
+    public DataBaseHandler(final File pluginDir, final String filename) throws ConnectException {
```
SantoshNC68: same as above.

src/main/java/com/ericsson/gerrit/plugins/eiffel/handlers/DataBaseHandler.java
```diff
@@ -0,0 +1,245 @@
...
+    }
+
+    /**
+     * This function will create a sql database structure in the database file if it
+     * does not exist.
+     *
+     * @throws ConnectException
+     */
+    private void createNewDatabase() throws ConnectException {
+        try (Connection sqlConnection = connect()) {
```
SantoshNC68: are we really creating a DB here?
I just see 2 calls one to close the connection and 2 to get the metadata. 
Is it not the connect() which actually creates DB?

src/main/java/com/ericsson/gerrit/plugins/eiffel/handlers/DataBaseHandler.java
```diff
@@ -0,0 +1,245 @@
...
+     * Creates a prepared statement and executes an update on the sqlStatement
+     * given.
+     *
+     * @param sqlStatement
+     * @param keyValue
+     * @param eiffelevent
+     * @throws ConnectException
+     * @throws SQLException
+     */
+    private void executeUpdate(final String sqlStatement, final String keyValue, final String eiffelevent)
```
SantoshNC68: eiffelEvent

src/main/java/com/ericsson/gerrit/plugins/eiffel/handlers/DataBaseHandler.java
```diff
@@ -0,0 +1,245 @@
...
+     * This function inserts values to the given table. The keyValue value is
+     * different depending on Table (branch name for scs and change-id for scc)
+     *
+     * @param table
+     * @param keyValue
+     * @param eiffelevent
+     * @throws SQLException
+     * @throws ConnectException
+     */
+    public void insertInto(final Table table, final String keyValue, final String eiffelevent)
```
SantoshNC68: eiffelEvent

src/main/java/com/ericsson/gerrit/plugins/eiffel/handlers/DataBaseHandler.java
```diff
@@ -0,0 +1,245 @@
...
+                throw new SQLException("No changes was saved in the database.");
+            }
+
+        } catch (SQLException e) {
+            LOGGER.error("Error when trying to add value into database: {}\n{}", e.getMessage(), e);
+            throw e;
+        }
+    }
+
+    public void createTable(Table table, Statement statement) throws ConnectException, SQLException {
```
SantoshNC68: the parameters can be final

src/main/java/com/ericsson/gerrit/plugins/eiffel/handlers/NoSuchElementException.java
```diff
@@ -0,0 +1,25 @@
...
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package com.ericsson.gerrit.plugins.eiffel.handlers;
+
+public class NoSuchElementException extends Exception {
```
SantoshNC68: should be moved to exception package

src/main/java/com/ericsson/gerrit/plugins/eiffel/state/SourceChangeCreatedState.java
```diff
@@ -0,0 +1,50 @@
...
+        super(pluginDir);
+    }
+
+    @Override
+    public String getEventId(String project, String changeId) throws NoSuchElementException, ConnectException, FileNotFoundException {
+        return getLastSubmittedEiffelEvent(project, changeId, Table.SCC_TABLE);
+    }
+
+    @Override
+    public void setState(String eiffelEventId, EiffelEvent eiffelEvent)
```
a-breid-ericsson: You still save the ID, not the entire event, this is also named eiffelEvent in DatabaseHandler, but should be ID, or actually maybe value as it could be something else in the future.

src/main/java/com/ericsson/gerrit/plugins/eiffel/state/StateFactory.java
```diff
@@ -0,0 +1,38 @@
...
+public class StateFactory {
+    public static State getState(File pluginDir, String eventType) {
+        try {
+            switch (eventType) {
+            case "EiffelSourceChangeCreatedEvent":
+                return new SourceChangeCreatedState(pluginDir);
+            case "EiffelSourceChangeSubmittedEvent":
+                return new SourceChangeSubmittedState(pluginDir);
+            }
+        } catch (Exception e) {
```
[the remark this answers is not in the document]
erik-edling: I wrote earlier that i will fix this things, its fixed in the upcoming change :)

src/main/java/com/ericsson/gerrit/plugins/eiffel/events/generators/EiffelEventGenerator.java
```diff
@@ -51,6 +63,33 @@ protected static String createRepoURI(String url, String projectName) {
         }
     }
 
+    protected static Link createLink(String linkType, String linkedEiffelEventType, File pluginDirectoryPath, final String projectName,
+            final String tableColumnName) {
```
m-linner-ericsson: Should we really know about tables when creating a link?

src/main/java/com/ericsson/gerrit/plugins/eiffel/storage/EventStorage.java
```diff
@@ -0,0 +1,147 @@
...
+        }
+    }
+
+    protected void setLastSubmittedEiffelEvent(String project, String tableColumnName,
+            String eiffelEvent,
+            Table tableName) throws NoSuchElementException, ConnectException, SQLException {
+        String fileName = String.format("%s.%s", project, FILE_ENDING);
+        DataBaseHandler dBHandler = new DataBaseHandler(pluginDir, fileName);
+        String parentPath = buildParentFilePath(pluginDir, project);
+        createParentDirsIfNecessary(parentPath);
```
m-linner-ericsson: You need to explain why you need the parent file path because now it looks you are just creating them without using them.

src/main/java/com/ericsson/gerrit/plugins/eiffel/handlers/DataBaseHandler.java
```diff
@@ -0,0 +1,231 @@
...
+     * @param table
+     * @param keyValue
+     * @return eventId
+     * @throws ConnectException
+     * @throws NoSuchElementException
+     */
+    public String getEventID(final Table table, final String keyValue) throws ConnectException, NoSuchElementException {
+        String eventID = "";
+
+        String sqlSelectStatement = String.format("SELECT * FROM %s WHERE %s=?", table, table.keyName);
```
m-linner-ericsson: I think the name of the function is confusing. We are not just returning an id but two values as we searching for "*".

src/main/java/com/ericsson/gerrit/plugins/eiffel/storage/EventStorage.java
```diff
@@ -0,0 +1,147 @@
...
+        } catch (NoSuchElementException e) {
+            LOGGER.debug("No previous event was saved, creating a new entry");
+            return "";
+        }
+    }
+
+    private String getEventId(String project, String tableColumnName,
+            Table tableName) throws NoSuchElementException, ConnectException {
+        String fileName = String.format("%s.%s", project, FILE_ENDING);
+        DataBaseHandler dBHandler = new DataBaseHandler(pluginDir, fileName);
```
m-linner-ericsson: This code is duplicated in two places. Why not initialize the dbHandler in the constructor?

src/main/java/com/ericsson/gerrit/plugins/eiffel/messaging/EiffelEventSender.java
```diff
@@ -89,12 +106,22 @@ public EiffelEventSender setEiffelEventType(final String eiffelType) {
...
         verifyResponse(response);
+
+        String generatedEventId = getGeneratedEventId(response);
+        return generatedEventId;
+    }
+
+    private String getGeneratedEventId(final ResponseEntity response) {
+        JsonObject responseBody = new JsonParser().parse(response.getBody()).getAsJsonObject();
+        JsonArray events = responseBody.get("events").getAsJsonArray();
+        String generatedEventId = events.get(0).getAsJsonObject().get("id").getAsString();
```
m-linner-ericsson: Are you certain that you allways have a "0"? This could cause a null pointer exception

src/main/java/com/ericsson/gerrit/plugins/eiffel/handlers/DataBaseHandler.java
```diff
@@ -0,0 +1,231 @@
...
+                throw new SQLException("No changes was saved in the database.");
+            }
+
+        } catch (SQLException e) {
+            LOGGER.error("Error when trying to add value into database: {}\n{}", e.getMessage(), e);
+            throw e;
+        }
+    }
+
+    public void createTable(final Table table, final Statement statement) throws ConnectException, SQLException {
```
m-linner-ericsson: Why public?

src/main/java/com/ericsson/gerrit/plugins/eiffel/handlers/MessageQueueHandler.java
```diff
@@ -17,15 +17,15 @@
...
 import com.google.gerrit.extensions.annotations.PluginName;
 import com.google.gerrit.extensions.events.LifecycleListener;
 import com.google.gerrit.server.config.PluginConfig;
 import com.google.gerrit.server.config.PluginConfigFactory;
 import com.google.gerrit.server.git.WorkQueue;
 import com.google.inject.Inject;
 
-import java.util.concurrent.ScheduledThreadPoolExecutor;
-
 public class MessageQueueHandler implements LifecycleListener {
```
m-linner-ericsson: Can you put a javadoc on how this is used since I can't find a place where we actually add something to the queue.

src/main/java/com/ericsson/gerrit/plugins/eiffel/storage/EventStorageFactory.java
```diff
@@ -0,0 +1,39 @@
...
+
+public class EventStorageFactory {
+    protected static final Logger LOGGER = LoggerFactory.getLogger(EventStorageFactory.class);
+
+    public static EventStorage getEventStorage(File pluginDir, String eventType) {
+        switch (eventType) {
+            case "EiffelSourceChangeCreatedEvent":
+                return new SourceChangeCreatedStorage(pluginDir);
+            case "EiffelSourceChangeSubmittedEvent":
+                return new SourceChangeSubmittedStorage(pluginDir);
```
m-linner-ericsson: Can't we use the `EventType`here?

src/main/java/com/ericsson/gerrit/plugins/eiffel/events/generators/EiffelEventGenerator.java
```diff
@@ -51,6 +63,40 @@ protected static String createRepoURI(String url, String projectName) {
...
+            return link;
+        }
+
+        } catch(IllegalArgumentException e) {
+            LOGGER.error("Failed creating link.", e);
+        }
+        return null;
+    }
+
+    protected static String getEiffelEventIdFromState(final EventStorage eventStorage, final String projectName,
```
m-linner-ericsson: I think you need to change the name to `getEiffelEventIdFromStorage` or something

src/main/java/com/ericsson/gerrit/plugins/eiffel/handlers/DataBaseHandler.java
m-linner-ericsson: But that should be explained

src/main/java/com/ericsson/gerrit/plugins/eiffel/events/generators/EiffelEventGenerator.java
erik-edling: Broke the event fetching into its own method in upcoming change :)

src/main/java/com/ericsson/gerrit/plugins/eiffel/handlers/DataBaseHandler.java
erik-edling: It's actually just returning one value, we only have one stored. Otherwise nothing would have worked.

src/main/java/com/ericsson/gerrit/plugins/eiffel/events/generators/EiffelEventGenerator.java
```diff
@@ -51,6 +63,33 @@ protected static String createRepoURI(String url, String projectName) {
         }
     }
 
+    protected static Link createLink(String linkType, String linkedEiffelEventType, File pluginDirectoryPath, final String projectName,
+            final String tableColumnName) {
+        State state = StateFactory.getState(pluginDirectoryPath, linkedEiffelEventType);
+        String lastEiffelEvent = getEiffelEventIdFromState(state, projectName, tableColumnName);
```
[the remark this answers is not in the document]
erik-edling: Done

src/main/java/com/ericsson/gerrit/plugins/eiffel/storage/EventStorage.java
erik-edling: This file is mostly taken from the old code and i actually don't understand this quite, i will check :)

src/main/java/com/ericsson/gerrit/plugins/eiffel/handlers/DataBaseHandler.java
erik-edling: I have no idéa, will change in upcoming change

src/main/java/com/ericsson/gerrit/plugins/eiffel/storage/EventStorageFactory.java
erik-edling: Ye, I've changed that!

src/main/java/com/ericsson/gerrit/plugins/eiffel/handlers/MessageQueueHandler.java
erik-edling: Maybe @a-breid-ericsson can do that in another PR as he wrote this? I don't understand this either.

src/main/java/com/ericsson/gerrit/plugins/eiffel/events/generators/EiffelEventGenerator.java
erik-edling: Ye, i missed that one, thanks 😄

src/main/java/com/ericsson/gerrit/plugins/eiffel/handlers/DataBaseHandler.java
erik-edling: Added a comment for that.

src/main/java/com/ericsson/gerrit/plugins/eiffel/events/generators/EiffelEventGenerator.java
```diff
@@ -51,6 +63,33 @@ protected static String createRepoURI(String url, String projectName) {
...
+
+        if (!StringUtils.isEmpty(lastEiffelEvent)) {
```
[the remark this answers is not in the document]
erik-edling: Fixed, it was the wrong import

src/main/java/com/ericsson/gerrit/plugins/eiffel/handlers/DataBaseHandler.java
m-linner-ericsson: Ok, `executeQuery` does the magic. Is it reasonable to update the query to fetch only the wanted value or am I too picky

src/test/java/com/ericsson/gerrit/plugins/eiffel/events/generators/EiffelEventGeneratorTest.java
erik-edling: Now the tests are in 😄

src/test/java/com/ericsson/gerrit/plugins/eiffel/storage/EventStorageTest.java
```diff
@@ -0,0 +1,115 @@
...
+
+    @Rule
+    public final ExpectedException exception = ExpectedException.none();
+
+
+    @Before
+    public void init() throws Exception {
+        tmpFolderPath = testFolder.newFolder();
+        dbHandler = Mockito.mock(DataBaseHandler.class);
+        PowerMockito.whenNew(DataBaseHandler.class).withParameterTypes(File.class, String.class).withArguments(Mockito.any())
```
a-breid-ericsson: Since you use both PowerMockito and Mockito it is good that you write out what you use. I like this, well done. :thumbsup:

src/test/java/com/ericsson/gerrit/plugins/eiffel/messaging/EiffelEventSenderTest.java
```diff
@@ -33,47 +34,49 @@ public void beforeTest() throws IOException, URISyntaxException {
         setUpMockObjects();
     }
 
-    @Test
-    public void testEventSender() throws Exception {
-        setUpMockActions();
+     @Test
```
a-breid-ericsson: Here is tab size 5 and not 4.

src/test/java/com/ericsson/gerrit/plugins/eiffel/handlers/DataBaseHandlerTest.java
```diff
@@ -0,0 +1,217 @@
...
+        // return an empty String,
+        // exception may be thrown when no values was found and should be empty.
+        if (result.next()) {
+            Mockito.when(result.next()).thenThrow(new SQLException("Exception thrown by test"));
+            exception.expect(RuntimeException.class);
+            dbHandler.getEventID(Table.SCS_TABLE, branch);
+        } else {
+            throw new NoSuchElementException("Mocking ResultSet class was unsuccessful...!");
+        }
+
```
a-breid-ericsson: :eyes: Since the test throws an exception at this point, and test except exceptions here, test will pass but code below will never be executed. To make it more clear please split the test into several tests so it is easier to see that each test passes.

src/test/java/com/ericsson/gerrit/plugins/eiffel/handlers/DataBaseHandlerTest.java
```diff
@@ -0,0 +1,217 @@
...
+import org.powermock.api.mockito.PowerMockito;
+import org.powermock.core.classloader.annotations.PrepareForTest;
+import org.powermock.modules.junit4.PowerMockRunner;
+
+import com.ericsson.gerrit.plugins.eiffel.exceptions.NoSuchElementException;
+
+@RunWith(PowerMockRunner.class)
+@PrepareForTest(fullyQualifiedNames = "com.ericsson.gerrit.plugins.eiffel.*")
+public class DataBaseHandlerTest {
+    private final String branch = "my_test_branch";
```
a-breid-ericsson: Please make those variables static and in uppercase so it is easier to see in the code where they come from. Please also have a calid_branch and in incorrect_branch. This makes it easier to undertsnd that a faulty branch gives no response.

src/main/java/com/ericsson/gerrit/plugins/eiffel/storage/EventStorage.java
a-breid-ericsson: Parent file path is when a project has parents, /somepath/baseProjectDir/children

It creates the dirs in local server disk so the DB file is saved in an existing directory, maybe this should be in the DBHandler. DBHandler will faiil to save the db file if the dirst does not exist.

src/main/java/com/ericsson/gerrit/plugins/eiffel/handlers/Table.java
```diff
@@ -0,0 +1,32 @@
...
+
+package com.ericsson.gerrit.plugins.eiffel.handlers;
+
+public enum Table {
+    SCS_TABLE("branch"), SCC_TABLE("changeId");
+
+    final String keyName;
+
+    Table(String keyValue) {
+        this.keyName = keyValue;
```
a-breid-ericsson: What if keyName was public final? then one could access it from my_table.keyName and not getKeyName? would that look better.

src/main/java/com/ericsson/gerrit/plugins/eiffel/handlers/MessageQueueHandler.java
m-linner-ericsson: Fixed myself in https://github.com/eiffel-community/eiffel-gerrit-plugin/pull/39

src/main/java/com/ericsson/gerrit/plugins/eiffel/handlers/DataBaseHandler.java
SantoshNC68: Ok for me

src/main/java/com/ericsson/gerrit/plugins/eiffel/storage/EventStorage.java
m-linner-ericsson: Sound like this should be moved then

src/test/java/com/ericsson/gerrit/plugins/eiffel/handlers/DataBaseHandlerTest.java
erik-edling: It feels weird to even do these if/else statements. If the mocking fails, the test should fail? Is it okay if i don't break these old tests up now?

src/test/java/com/ericsson/gerrit/plugins/eiffel/handlers/DataBaseHandlerTest.java
erik-edling: done, i might have get you wrong though

src/main/java/com/ericsson/gerrit/plugins/eiffel/handlers/Table.java
erik-edling: I'm sorry i couldnt take a look at it today and i won't be here tomorrow. I will ask @saif-ericsson to look into this :)

src/main/java/com/ericsson/gerrit/plugins/eiffel/messaging/EiffelEventSender.java
```diff
@@ -67,20 +79,25 @@ public EiffelEventSender(EiffelPluginConfiguration pluginConfig, HttpRequest htt
...
         } catch (HttpRequestFailedException e) {
             LOGGER.error("Failed to send eiffel message.", e);
             throw e;
+        } catch (IOException e) {
+            LOGGER.error("Failed to send eiffel message.", e);
+            throw new HttpRequestFailedException(e);
         }
     }
 
     public EiffelEventSender setEiffelEventMessage(final EiffelEvent eiffelMessage) {
```
SantoshNC68: change to eiffelEvent to maintain parameters consistency

src/main/java/com/ericsson/gerrit/plugins/eiffel/events/generators/EiffelEventGenerator.java
```diff
@@ -51,6 +63,41 @@ protected static String createRepoURI(String url, String projectName) {
         }
     }
 
+    protected static String getPreviousEiffelEvent(String linkedEiffelEventType, String projectName, String searchCriteria, File pluginDirectoryPath) {
+        try {
+            EventStorage eventStorage = EventStorageFactory.getEventStorage(pluginDirectoryPath, linkedEiffelEventType);
+            String lastEiffelEvent = getEiffelEventIdFromStorage(eventStorage, projectName, searchCriteria);
+            return lastEiffelEvent;
+        } catch(IllegalArgumentException e) {
+            LOGGER.error("Failed creating link.", e);
```
m-linner-ericsson: Is this error message correct or something left?

src/main/java/com/ericsson/gerrit/plugins/eiffel/storage/EventStorage.java
```diff
@@ -0,0 +1,98 @@
...
+                    searchCriteria);
+        }
+    }
+
+    private String getOldEventId(DatabaseHandler dBHandler, Table tableName, String searchCriteria) throws ConnectException {
+        try {
+            String oldEvent = dBHandler.getEventID(tableName, searchCriteria);
+            return oldEvent;
+        } catch (NoSuchElementException e) {
+            LOGGER.debug("No previous event was saved, creating a new entry");
```
m-linner-ericsson: I would like to rephrase this error message to "No old event found" or something. No new entry is created in this method.

src/main/java/com/ericsson/gerrit/plugins/eiffel/storage/EventStorage.java
```diff
@@ -0,0 +1,98 @@
...
+
+    protected String getLastSavedEiffelEvent(String project, String searchCriteria,
+            Table tableName)
+            throws NoSuchElementException, FileNotFoundException, ConnectException {
+        try {
+            return getEventId(project, searchCriteria, tableName);
+        } catch (NoSuchElementException e) {
+            throw new NoSuchElementException(
+                    "Database did not return any value for this query\n" + "Exception Message:" + e.getMessage());
+        }
```
m-linner-ericsson: What is the benefit of catching and rethrowing the same exception? I don't feel this text adds any information.

src/main/java/com/ericsson/gerrit/plugins/eiffel/handlers/DataBaseHandler.java
```diff
@@ -0,0 +1,232 @@
...
+     * @param keyValue
+     * @param value
+     * @throws SQLException
+     * @throws ConnectException
+     */
+    public void insertInto(final Table table, final String keyValue, final String value)
+            throws SQLException, ConnectException {
+        String sqlInsertStatement = String.format("INSERT INTO %s(%s,%s) VALUES(?,?)", table, EVENT_ID_KEY,
+                table.keyName);
+        executeUpdate(sqlInsertStatement, keyValue, value);
```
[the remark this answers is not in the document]
m-linner-ericsson: No this can only insert not update

src/main/java/com/ericsson/gerrit/plugins/eiffel/exceptions/NoSuchElementException.java
```diff
@@ -0,0 +1,25 @@
...
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package com.ericsson.gerrit.plugins.eiffel.exceptions;
+
+public class NoSuchElementException extends Exception {
```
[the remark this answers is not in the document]
a-breid-ericsson: Did we not agree not to have Java doc in exceptions? Not sure anyone using this would have help from java doc. But if other exceptions has, this one should as well.

src/test/java/com/ericsson/gerrit/plugins/eiffel/handlers/DataBaseHandlerTest.java
a-breid-ericsson: Your call but if not now it will never happen.

src/test/java/com/ericsson/gerrit/plugins/eiffel/handlers/DataBaseHandlerTest.java
a-breid-ericsson: Much better as static.
If you feel for it you could ad a specific static for branch name if the test is supposed to fail, just to make it easier to read. But already better as static since you now see where the variable comes from.

src/main/java/com/ericsson/gerrit/plugins/eiffel/events/generators/EiffelEventGenerator.java
```diff
@@ -51,6 +63,41 @@ protected static String createRepoURI(String url, String projectName) {
         }
     }
 
+    protected static String getPreviousEiffelEvent(String linkedEiffelEventType, String projectName, String tableColumnName, File pluginDirectoryPath) {
```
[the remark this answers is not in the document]
saif-ericsson: Well, I changed it But for small methods it doesnt make much sense, though good programming practice.

src/main/java/com/ericsson/gerrit/plugins/eiffel/handlers/DataBaseHandler.java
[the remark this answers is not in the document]
SantoshNC68: executeUpdate(sqlInsertStatement, keyValue, value); this method is reference both in insertInto and updateInto. That is where it causes confusion.
I would rather have the name to execute as it does nothing more than executing some statements.

src/main/java/com/ericsson/gerrit/plugins/eiffel/events/generators/EiffelEventGenerator.java
saif-ericsson: Not sure... any idea?

src/main/java/com/ericsson/gerrit/plugins/eiffel/storage/EventStorage.java
saif-ericsson: "No old event found"

src/main/java/com/ericsson/gerrit/plugins/eiffel/handlers/Table.java
saif-ericsson: I am leaving it as it is... as it doesn't make any difference.

src/test/java/com/ericsson/gerrit/plugins/eiffel/handlers/DataBaseHandlerTest.java
saif-ericsson: If/else were just meant to satisfy codacy. The problem was "else" which throw the exception. Now I try to fix it by mocking result.next(). You may come up with some other solutions.

src/main/java/com/ericsson/gerrit/plugins/eiffel/handlers/DataBaseHandler.java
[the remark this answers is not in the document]
saif-ericsson: Changed it to "prepareAndExecuteStatement"

src/main/java/com/ericsson/gerrit/plugins/eiffel/events/generators/EiffelEventGenerator.java
m-linner-ericsson: Could not get previous eiffel event

src/main/java/com/ericsson/gerrit/plugins/eiffel/messaging/EiffelEventSender.java
```diff
@@ -89,12 +106,23 @@ public EiffelEventSender setEiffelEventType(final String eiffelType) {
...
+
+        String generatedEventId = getGeneratedEventId(response);
+        return generatedEventId;
+    }
+
+    private String getGeneratedEventId(final ResponseEntity response) {
+        final String jsonBody = response.getBody();
+        final JsonObject responseBody = new JsonParser().parse(jsonBody).getAsJsonObject();
+        final JsonArray events = responseBody.get("events").getAsJsonArray();
+        final String generatedEventId = events.get(0).getAsJsonObject().get("id").getAsString();
```
m-linner-ericsson: Are you certain that you always have a "0" or an "id"? This could cause a null pointer exception.

erik-edling: As long as we are getting 200 it will be this structure. If we get something else, we should maybe throw an exception and never get to this place.

Christoffer-Cortes: I think it may be fine the way it is. verifyResponse will check the status code and return an exception if it is 500 or 503. This is to make sure it retries though so for any other status code no exception is thrown but I guess this could be expanded in there somehow.

src/main/java/com/ericsson/gerrit/plugins/eiffel/handlers/DatabaseHandler.java
```diff
@@ -0,0 +1,273 @@
...
+                String parentPath = buildParentFilePath();
+                createParentDirsIfNecessary(parentPath);
+
+                //When getMetaData is called, it will create the .db file
+                sqlConnection.getMetaData();
+                sqlConnection.close();
+            }
+
+        } catch (SQLException e) {
+            LOGGER.error("Error when trying to create new database: {}\n{}", e.getMessage(), e);
```
Christoffer-Cortes: I believe e contains the message if printed. You don't need to add it with brackets either. just adding e as a parameter is enough to have in printed. Ex: ("Error when trying to create new database.", e)

src/main/java/com/ericsson/gerrit/plugins/eiffel/handlers/DatabaseHandler.java
```diff
@@ -0,0 +1,273 @@
...
+     * @throws ConnectException
+     */
+    private void createTables() throws ConnectException {
+        try (Connection connection = connect(); Statement statement = connection.createStatement()) {
+            for (Table table : Table.values()) {
+                createTable(table, statement);
+            }
+            LOGGER.debug("Created tables successfully");
+        } catch (SQLException e) {
+            LOGGER.error("Error while creating Tables in database: {}\n{}", e.getMessage(), e);
```
Christoffer-Cortes: Same as above.

src/main/java/com/ericsson/gerrit/plugins/eiffel/storage/EventStorage.java
erik-edling: No value here, we removed it ;)

src/main/java/com/ericsson/gerrit/plugins/eiffel/handlers/DatabaseHandler.java
```diff
@@ -0,0 +1,273 @@
...
+     * @return Connection to database
+     * @throws ConnectException
+     */
+    private Connection connect() throws ConnectException {
+        try {
+            Class.forName("org.sqlite.JDBC");
+            return DriverManager.getConnection(this.databaseFile);
+        } catch (SQLException | ClassNotFoundException e) {
+            throw new ConnectException(
+                    String.format("Failed to create a database connection. %s\n%s", e.getMessage(), e));
```
Christoffer-Cortes: This I also believe to be wrong, Exceptions usually support exception as a parameter. This could then be written as: throw new ConnectException("Failed to create a database connection.", e);

src/main/java/com/ericsson/gerrit/plugins/eiffel/handlers/DatabaseHandler.java
erik-edling: Agreed :)

src/main/java/com/ericsson/gerrit/plugins/eiffel/events/generators/EiffelEventGenerator.java
```diff
@@ -51,6 +63,41 @@ protected static String createRepoURI(String url, String projectName) {
...
+            Link link = new Link();
+            link.type = linkType;
+            link.target = lastEiffelEvent;
+
+            return link;
+        }
+        return null;
+    }
+
+    protected static String getEiffelEventIdFromStorage(final EventStorage eventStorage, final String projectName,
```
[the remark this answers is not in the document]
saif-ericsson: We have made it private and thus they have moved further apart.

src/main/java/com/ericsson/gerrit/plugins/eiffel/handlers/DataBaseHandler.java
```diff
@@ -0,0 +1,232 @@
...
+            preparedStatement.setString(1, eiffelEvent);
+            preparedStatement.setString(2, keyValue);
+            int updateCount = preparedStatement.executeUpdate();
+
+            if (updateCount == 0) {
+                throw new SQLException("No changes was saved in the database.");
+            }
+
+        } catch (SQLException e) {
+            LOGGER.error("Error when trying to add value into database: {}\n{}", e.getMessage(), e);
```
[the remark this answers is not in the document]
saif-ericsson: done

src/test/java/com/ericsson/gerrit/plugins/eiffel/handlers/DataBaseHandlerTest.java
a-breid-ericsson: The problem here is that not all parts of the test is run since the NoSuchElementException is thrown before the below tests. This test needs to be split intoo several tests and test every exception on it´s own.

src/main/java/com/ericsson/gerrit/plugins/eiffel/handlers/DataBaseHandler.java
```diff
@@ -0,0 +1,232 @@
...
+     * This function returns an event id if exists for a specific table depending on
+     * the keyValue
+     *
+     * @param table
+     * @param keyValue
+     * @return eventId
+     * @throws ConnectException
+     * @throws NoSuchElementException
+     */
+    public String getEventID(final Table table, final String keyValue) throws ConnectException, NoSuchElementException {
```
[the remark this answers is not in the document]
saif-ericsson: that was outdated code

src/main/java/com/ericsson/gerrit/plugins/eiffel/handlers/DataBaseHandler.java
```diff
@@ -0,0 +1,232 @@
...
+     * This function updates value to the given table. The keyValue value is
+     * different depending on Table (branch name for scs and change-id for scc)
+     *
+     * @param table
+     * @param keyValue
+     * @param eiffelEvent
+     * @throws ConnectException
+     * @throws SQLException
+     */
+    public void updateInto(final Table table, final String keyValue, final String eiffelEvent)
```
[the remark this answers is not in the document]
saif-ericsson: that was outdated code

src/main/java/com/ericsson/gerrit/plugins/eiffel/events/generators/EiffelEventGenerator.java
```diff
@@ -63,18 +63,19 @@ protected static String createRepoURI(String url, String projectName) {
...
-    protected static String getPreviousEiffelEvent(String linkedEiffelEventType, String projectName, String searchCriteria, File pluginDirectoryPath) {
+    protected static String getPreviousEiffelEvent(final String linkedEiffelEventType, final String projectName,
+            final String searchCriteria, final File pluginDirectoryPath) {
         try {
             EventStorage eventStorage = EventStorageFactory.getEventStorage(pluginDirectoryPath, linkedEiffelEventType);
             String lastEiffelEvent = getEiffelEventIdFromStorage(eventStorage, projectName, searchCriteria);
             return lastEiffelEvent;
         } catch(IllegalArgumentException e) {
-            LOGGER.error("Failed creating link.", e);
+            LOGGER.error("Could not get previous eiffel event.", e);
```
e-pettersson-ericsson: eiffel with capital E

src/main/java/com/ericsson/gerrit/plugins/eiffel/events/generators/EiffelEventGenerator.java
```diff
@@ -85,22 +86,23 @@ protected static Link createLink(final String linkType, String lastEiffelEvent)
...
+        return uri.toString();
+    }
+
+    private static String getEiffelEventIdFromStorage(final EventStorage eventStorage, final String projectName,
+            final String searchCriteria) {
         try {
             String eventId = eventStorage.getEventId(projectName, searchCriteria);
             return eventId;
         } catch (NoSuchElementException e) {
+            LOGGER.debug("Event Storage didnt return any value for this query.", e);
```
e-pettersson-ericsson: didnt  --> did not

src/main/java/com/ericsson/gerrit/plugins/eiffel/events/generators/EiffelEventGenerator.java
saif-ericsson: BIG E

src/main/java/com/ericsson/gerrit/plugins/eiffel/handlers/DatabaseHandler.java
```diff
@@ -0,0 +1,275 @@
...
+ * exist as a db file, if not it gets created together with required tables. At
+ * initiation a file name is given making it possible to store a file per
+ * project depending how the classes that uses this handler implements this
+ * function.
+ *
+ */
+public class DatabaseHandler {
+    private static final Logger LOGGER = LoggerFactory.getLogger(DatabaseHandler.class);
+    private static final String EVENT_ID_KEY = "eventId";
+    protected static final String FILE_ENDING = "db";
```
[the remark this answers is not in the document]
saif-ericsson: Now Typed

src/main/java/com/ericsson/gerrit/plugins/eiffel/events/generators/EiffelEventGenerator.java
```diff
@@ -51,9 +63,46 @@ protected static String createRepoURI(String url, String projectName) {
...
+    private static String getSshBaseUrl(final String host) throws URISyntaxException {
         URI uri;
         uri = new URI("ssh", null, host, DEFAULT_SSH_PORT, "/", null, null);
         return uri.toString();
     }
+
+    private static String getEiffelEventIdFromStorage(final EventStorage eventStorage, final String projectName,
+            final String searchCriteria) {
+        try {
+            String eventId = eventStorage.getEventId(projectName, searchCriteria);
```
SantoshNC68: may be return without having to add to a string something like
return eventStorage.getEventId(...)

src/main/java/com/ericsson/gerrit/plugins/eiffel/events/generators/EiffelSourceChangeCreatedEventGenerator.java
```diff
@@ -71,15 +78,17 @@ public static EiffelSourceChangeCreatedEvent generate(PatchSetCreatedEvent patch
...
-        // getLatestEiffelSourceChangeSubmittedEventId();
-        // setBaseLink(latestEiffelSourceChangeSubmittedEventId);
+        String previousSourceChangeCreatedEvent = getPreviousEiffelEvent(EventType.SCC_EVENT, projectName, changeId, pluginDirectoryPath);
+        final Link previousVersionLink = createLink(LINK_TYPE_PREVIOUS_VERSION, previousSourceChangeCreatedEvent);
+        if (previousVersionLink != null) {
+            eiffelEvent.eventParams.links.add(previousVersionLink);
+        }
 
+        String previousSourceChangeSubmittedEvent = getPreviousEiffelEvent(EventType.SCS_EVENT, projectName, changeId, pluginDirectoryPath);
+        final Link baseLink = createLink(LINK_TYPE_BASE, previousSourceChangeSubmittedEvent);
```
SantoshNC68: can we move this also to the function createLink?
like createLinkAndAddtoEvent(type, event, eiffelEvent) in that case we can avoid the null check and duplication of code moved to the function.

src/main/java/com/ericsson/gerrit/plugins/eiffel/events/generators/EiffelSourceChangeCreatedEventGenerator.java
```diff
@@ -71,15 +78,17 @@ public static EiffelSourceChangeCreatedEvent generate(PatchSetCreatedEvent patch
...
-        // String latestEiffelSourceChangeSubmittedEventId =
-        // getLatestEiffelSourceChangeSubmittedEventId();
-        // setBaseLink(latestEiffelSourceChangeSubmittedEventId);
+        String previousSourceChangeCreatedEvent = getPreviousEiffelEvent(EventType.SCC_EVENT, projectName, changeId, pluginDirectoryPath);
+        final Link previousVersionLink = createLink(LINK_TYPE_PREVIOUS_VERSION, previousSourceChangeCreatedEvent);
+        if (previousVersionLink != null) {
+            eiffelEvent.eventParams.links.add(previousVersionLink);
+        }
 
+        String previousSourceChangeSubmittedEvent = getPreviousEiffelEvent(EventType.SCS_EVENT, projectName, changeId, pluginDirectoryPath);
```
m-linner-ericsson: This can't be tested properly. If you try to find the SCS event based on the changeId you will always get an empty result

src/main/java/com/ericsson/gerrit/plugins/eiffel/events/generators/EiffelEventGenerator.java
erik-edling: I think that's less readable. Better adding to string before returning

src/main/java/com/ericsson/gerrit/plugins/eiffel/events/generators/EiffelSourceChangeCreatedEventGenerator.java
erik-edling: Then that method will do multiple things plus it feels wrong sending in a reference and update it from readability point of view. It doesnt feel as much readable that way.

src/main/java/com/ericsson/gerrit/plugins/eiffel/events/generators/EiffelSourceChangeCreatedEventGenerator.java
erik-edling: As we talked about, in unittest. We won't test this since that functionality is broken apart with (this class, storage class and db handled). We will try to make sure that we are sending the correct values to each method though and later, functional tests will include this functionality.

src/test/java/com/ericsson/gerrit/plugins/eiffel/events/generators/EiffelEventGeneratorTest.java
```diff
@@ -212,8 +240,11 @@ private void verifyEiffelSourceChangeSubmittedEvent(
         assertEquals(errorMessage, COMMIT_ID, gitIdentifier.get("commitId").getAsString());
         assertEquals(errorMessage, BRANCH, gitIdentifier.get("branch").getAsString());
         assertEquals(errorMessage, PROJECT, gitIdentifier.get("repoName").getAsString());
+        assertEquals(errorMessage, "CHANGE", link1.get("type").getAsString());
+        assertEquals(errorMessage, "PREVIOUS_VERSION", link2.get("type").getAsString());
```
m-linner-ericsson: There are 4 cases but you are only testing 1 case

src/test/java/com/ericsson/gerrit/plugins/eiffel/events/generators/EiffelEventGeneratorTest.java
```diff
@@ -240,5 +276,7 @@ private void verifyEiffelSourceChangeCreatedEvent(
         assertEquals(errorMessage, CHANGE_ID, change.get("id").getAsString());
         assertEquals(errorMessage, SIZE_INSERTIONS, change.get("insertions").getAsInt());
         assertEquals(errorMessage, SIZE_DELETIONS, change.get("deletions").getAsInt());
+        assertEquals(errorMessage, "PREVIOUS_VERSION", link1.get("type").getAsString());
+        assertEquals(errorMessage, "BASE", link2.get("type").getAsString());
```
m-linner-ericsson: There are 4 cases but you are only testing 1 case

src/test/java/com/ericsson/gerrit/plugins/eiffel/events/generators/EiffelEventGeneratorTest.java
saif-ericsson: We are now covering: no links and link population in both events in the tests.

src/main/java/com/ericsson/gerrit/plugins/eiffel/events/generators/EiffelSourceChangeCreatedEventGenerator.java
saif-ericsson: We have added a test to check that we are sending the correct searchCriteria.

src/test/java/com/ericsson/gerrit/plugins/eiffel/events/generators/EiffelEventGeneratorTest.java
saif-ericsson: We are now covering: no links and link population in both events in the tests.

src/test/java/com/ericsson/gerrit/plugins/eiffel/handlers/DataBaseHandlerTest.java
saif-ericsson: next() is already mocked. We need to do this, otherwise codacy complains about the resultset.

src/test/java/com/ericsson/gerrit/plugins/eiffel/handlers/DataBaseHandlerTest.java
```diff
@@ -0,0 +1,212 @@
...
+        Mockito.doReturn(preparedStatement).when(connection).prepareStatement(Mockito.any());
+        Mockito.doReturn(result).when(preparedStatement).executeQuery();
+
+        // When mocking an exception while executing getEventID the function should
+        // return an empty String,
+        // exception may be thrown when no values was found and should be empty.
+        Mockito.when(result.next()).thenThrow(new SQLException("Exception thrown by test"));
+
+        exception.expect(RuntimeException.class);
+        dbHandler.getEventID(Table.SCS_TABLE, BRANCH);
```
a-breid-ericsson: Tetsed this.
The test concider it self completed at line 166 and nothing below is tested.

Issue #36, opened 2019-10-29
Perform request to REMReM using a queue


Christoffer-Cortes:
### Description
Related to issue https://github.com/eiffel-community/eiffel-gerrit-plugin/issues/31
Requests are not async at the moment and we need some sort of solution for this.

### Motivation
Synchronous requests to REMReM will block the listener thread. It needs to start an independent job and release the thread.

### Exemplification
https://github.com/eiffel-community/eiffel-gerrit-plugin/pull/32#issuecomment-545358994
RabbitMQ plugin mentioned here has an implementation that could be used.

### Benefits
Prevent threads from being held.

PR #38, opened 2019-10-30
Send calls are now threaded

Christoffer-Cortes:
### Applicable Issues
Closes https://github.com/eiffel-community/eiffel-gerrit-plugin/issues/36

### Description of the Change
Added a ThreadPoolExecutor that takes care of send calls.

### Alternate Designs
There are other alternative implementations in for example the rabbitmq plugin for gerrit and in some internal code.

### Benefits
EventListener threads are no longer held.

### Possible Drawbacks
The queue is set to have 3 workers at the moment and a queue of an arbitrary length. I guess this could lead to our plugin taking up more memory.

### Sign-off

    indicated in the file; or

    in the file; or

    it.

Signed-off-by: @Christoffer-Cortes

### DIFF
--- a/src/main/java/com/ericsson/gerrit/plugins/eiffel/listeners/AbstractEventListener.java
+++ b/src/main/java/com/ericsson/gerrit/plugins/eiffel/listeners/AbstractEventListener.java
@@ -17,6 +17,11 @@
 package com.ericsson.gerrit.plugins.eiffel.listeners;
 
 import java.io.File;
+import java.util.concurrent.ArrayBlockingQueue;
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.ThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 import org.apache.commons.lang.StringUtils;
 import org.slf4j.Logger;
@@ -47,6 +52,17 @@ public abstract class AbstractEventListener implements EventListener {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(AbstractEventListener.class);
 
+    private static final int CORE_POOL_SIZE = 3;
+    private static final int MAXIMUM_POOL_SIZE = 3;
+    private static final long KEEP_ALIVE = 0;
+    private static final TimeUnit TIME_UNIT = TimeUnit.MILLISECONDS;
+    private static final int QUEUE_SIZE = 16384;
+
+    private static AtomicBoolean initialized = new AtomicBoolean();
+    private static ThreadPoolExecutor executor;
+
+    private final BlockingQueue<Runnable> queue = new ArrayBlockingQueue<Runnable>(QUEUE_SIZE);
+
     @Inject
     private com.google.gerrit.server.config.PluginConfigFactory pluginConfigFactory;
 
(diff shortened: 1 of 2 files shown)


Christoffer-Cortes:
@sselberg @magnusbaeck I looked a bit at the rabbitmq code and saw that a BlockingQueue is created that stores events. This queue is then worked by a thread that publishes messages from this queue. I approached this issue by using a ThreadPoolExecutor. It also keeps a queue and can be set up with a number of workers. You then submit whatever piece of code you want to queue up. In our case that is the retry runnable.

magnusbaeck:
ThreadPoolExecutor SGTM. As long as the caller isn't blocked and the queue size is reasonably large it's probably the best thing we can do.

src/main/java/com/ericsson/gerrit/plugins/eiffel/listeners/AbstractEventListener.java
```diff
@@ -146,6 +164,13 @@ protected boolean isEiffelEventSendingEnabled(final Event gerritEvent,
     protected abstract void prepareAndSendEiffelEvent(Event gerritEvent,
             EiffelPluginConfiguration pluginConfig);
 
+    private void initializeThreadPoolExecutor() {
+        if (initialized.compareAndSet(false, true)) {
+            executor = new ThreadPoolExecutor(CORE_POOL_SIZE, MAXIMUM_POOL_SIZE, KEEP_ALIVE,
```
m-linner-ericsson: Doesn't this executor have to instantiated by injection? I can't see how this would continue to operate after we have exited the thread or am I missing something?

Christoffer-Cortes: The executor is static so its shared among any instance of the abstract listener which is extended by the other listeners. The executor itself is not a thread but is like a "worker" manager. When you submit something to the queue one of the available thread slots will be taken. In this case I've set it to a max of 3 simultaneous threads active.

m-linner-ericsson: So the thread that is started in this thread willl not be killed when this thread dies?

a-breid-ericsson: As I undertsnd it it is the Executor class that starts the thread and Gerrit is responsible of the executor class.

src/main/java/com/ericsson/gerrit/plugins/eiffel/listeners/AbstractEventListener.java
```diff
@@ -47,6 +52,17 @@
 
     private static final Logger LOGGER = LoggerFactory.getLogger(AbstractEventListener.class);
 
+    private static final int CORE_POOL_SIZE = 3;
+    private static final int MAXIMUM_POOL_SIZE = 3;
+    private static final long KEEP_ALIVE = 0;
+    private static final TimeUnit TIME_UNIT = TimeUnit.MILLISECONDS;
+    private static final int QUEUE_SIZE = 16384;
```
saif-ericsson: I m not sure about the logic of selecting these values, specially the queue sie of 16384 (may be 16kb)

Christoffer-Cortes: It's just an arbitrary choice for now. The queue size I took from rabbitmq plugin but it's also an arbitrary number as far as I know. It's half the size of a signed short.

PR #48, opened 2020-01-17
Link generator

m-linner-ericsson:
### Description of the Change
Move the code for generating links to its own class

### Alternate Designs
Keep the old code

### Benefits
More readable code

### Sign-off

    indicated in the file; or

    in the file; or

    it.

Signed-off-by: Mattias Linnér mattias.linner@ericsson.com

### DIFF
--- a/src/main/java/com/ericsson/gerrit/plugins/eiffel/events/generators/EiffelEventGenerator.java
+++ b/src/main/java/com/ericsson/gerrit/plugins/eiffel/events/generators/EiffelEventGenerator.java
@@ -20,25 +20,12 @@ import java.net.InetAddress;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.UnknownHostException;
-import java.util.List;
-
-import org.apache.commons.lang3.StringUtils;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
-import com.ericsson.gerrit.plugins.eiffel.configuration.EiffelPluginConfiguration;
-import com.ericsson.gerrit.plugins.eiffel.events.models.Link;
-import com.ericsson.gerrit.plugins.eiffel.exceptions.NoSuchElementException;
-import com.ericsson.gerrit.plugins.eiffel.storage.EventStorage;
-import com.ericsson.gerrit.plugins.eiffel.storage.EventStorageFactory;
 
 /**
  * Base class with common functionality for event generators.
  *
  */
 public class EiffelEventGenerator {
-    private static final Logger LOGGER = LoggerFactory.getLogger(EiffelEventGenerator.class);
-
     protected static final String META_SOURCE_NAME = "Eiffel Gerrit Plugin";
     private static final int DEFAULT_SSH_PORT = 29418;
 
@@ -64,48 +51,6 @@ public class EiffelEventGenerator {
         }
     }
 
-    /**
-     * Will for a given list of search criteria return the first found event
-     */
-    protected static String getPreviousEiffelEventId(final EiffelPluginConfiguration pluginConfig,
-            final String linkedEiffelEventType,
-            final List<String> searchCriterias) {
-        for (final String searchCriteria : searchCriterias) {
-            final String eiffelEventId = getPreviousEiffelEventId(pluginConfig,
-                    linkedEiffelEventType,
-                    searchCriteria);
-            if (!StringUtils.isEmpty(eiffelEventId)) {
-                return eiffelEventId;
-            }
-        }
-        return "";
-    }
-
-    protected static String getPreviousEiffelEventId(final EiffelPluginConfiguration pluginConfig,
-            final String linkedEiffelEventType,
-            final String searchCriteria ) {
-        try {
-            final EventStorage eventStorage = EventStorageFactory.getEventStorage(pluginConfig,
-                    linkedEiffelEventType);
-            final String lastEiffelEventId = getEiffelEventIdFromStorage(eventStorage, pluginConfig.getProject(),
-                    searchCriteria);
-            return lastEiffelEventId;
-        } catch (final IllegalArgumentException e) {
-            LOGGER.error("Could not get previous Eiffel event.", e);
-            return "";
-        }
-    }
-
-    protected static Link createLink(final String linkType, final String lastEiffelEvent) {
-        if (!StringUtils.isEmpty(lastEiffelEvent)) {
-            final Link link = new Link();
-            link.type = linkType;
-            link.target = lastEiffelEvent;
-
-            return link;
-        }
-        return null;
-    }
 
     private static String getSshBaseUrl(final String host) throws URISyntaxException {
         URI uri;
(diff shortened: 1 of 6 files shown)


src/main/java/com/ericsson/gerrit/plugins/eiffel/events/generators/EiffelEventGenerator.java
```diff
@@ -20,25 +20,12 @@
...
-import com.ericsson.gerrit.plugins.eiffel.events.models.Link;
-import com.ericsson.gerrit.plugins.eiffel.exceptions.NoSuchElementException;
-import com.ericsson.gerrit.plugins.eiffel.storage.EventStorage;
-import com.ericsson.gerrit.plugins.eiffel.storage.EventStorageFactory;
 
 /**
  * Base class with common functionality for event generators.
  *
  */
 public class EiffelEventGenerator {
```
SantoshNC68: Should this class be really called EiffelEventGenerator, as the code doesn't contain anything regarding eiffel events at al ?

SantoshNC68: Perhaps the name EiffelEventSourceGenerator is more apt

src/main/java/com/ericsson/gerrit/plugins/eiffel/events/generators/EiffelSourceChangeCreatedEventGenerator.java
```diff
@@ -80,21 +75,11 @@ public static EiffelSourceChangeCreatedEvent generate(final EiffelPluginConfigur
...
 
-        final String previousSourceChangeCreatedEvent = getPreviousEiffelEventId(pluginConfig,
-                EventType.SCC_EVENT, changeId);
-        final Link previousVersionLink = createLink(LINK_TYPE_PREVIOUS_VERSION,
-                previousSourceChangeCreatedEvent);
-        if (previousVersionLink != null) {
-            eiffelEvent.eventParams.links.add(previousVersionLink);
-        }
+        final LinkGenerator linkGenerator = new LinkGenerator(pluginConfig, commitInformation);
+        linkGenerator.addSccBase(commitId);
```
SantoshNC68: can't we move this logic to LinkGenerator directly? without having to explicitly call the addSccBase and addSccPreviousVersion?

src/main/java/com/ericsson/gerrit/plugins/eiffel/events/generators/EiffelSourceChangeSubmittedEventGenerator.java
```diff
@@ -70,22 +64,10 @@ public static EiffelSourceChangeSubmittedEvent generate(final EiffelPluginConfig
...
-        final String previousSourceChangeSubmittedEventId = getPreviousEiffelEventId(pluginConfig,
-                EventType.SCS_EVENT,
-                parentsSHAs);
-        final Link previousVersionLink = createLink(LINK_TYPE_PREVIOUS_VERSION,
-                previousSourceChangeSubmittedEventId);
-        if (previousVersionLink != null) {
-            eiffelEvent.eventParams.links.add(previousVersionLink);
-        }
+        final LinkGenerator linkGenerator = new LinkGenerator(pluginConfig, commitInformation);
+        linkGenerator.addScsChange(changeId);
```
SantoshNC68: If we move this logic to the constructor then we can avoid having to call it in every event.

src/main/java/com/ericsson/gerrit/plugins/eiffel/events/generators/EiffelEventGenerator.java
m-linner-ericsson: Not my naming but this is the base class for all the generators

src/main/java/com/ericsson/gerrit/plugins/eiffel/events/generators/EiffelSourceChangeCreatedEventGenerator.java
m-linner-ericsson: The links are different for the two events and I thought it was good to let that logic be part of the generator classes

src/main/java/com/ericsson/gerrit/plugins/eiffel/events/generators/EiffelSourceChangeSubmittedEventGenerator.java
m-linner-ericsson: What constructor?

Christoffer-Cortes: The two events have different links though so the constructor would not work for both. There would also be an increase of parameters passed.

src/main/java/com/ericsson/gerrit/plugins/eiffel/events/generators/LinkGenerator.java
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    private String getPreviousEiffelEventId(final String linkedEiffelEventType,
+            final String searchCriteria) {
+        try {
+            final EventStorage eventStorage = EventStorageFactory.getEventStorage(
+                    pluginConfig, linkedEiffelEventType);
+            return eventStorage.getEventId(pluginConfig.getProject(), searchCriteria);
+        } catch (final IllegalArgumentException e) {
+            LOGGER.error("Could not get previous Eiffel event.", e);
```
a-breid-ericsson: This could be a single catch and the message logged could be from the thrown exception.

src/main/java/com/ericsson/gerrit/plugins/eiffel/events/generators/EiffelEventGenerator.java
SantoshNC68: True, but it doesn't contain any logic pertaining to event generation. It basically extracts information from gerrit.

src/main/java/com/ericsson/gerrit/plugins/eiffel/events/generators/EiffelSourceChangeCreatedEventGenerator.java
SantoshNC68: ok.

src/main/java/com/ericsson/gerrit/plugins/eiffel/events/generators/EiffelSourceChangeSubmittedEventGenerator.java
SantoshNC68: ok

src/main/java/com/ericsson/gerrit/plugins/eiffel/events/generators/EiffelEventGenerator.java
m-linner-ericsson: I won't change the name in this pull request. Want to keep it small.

src/main/java/com/ericsson/gerrit/plugins/eiffel/events/generators/LinkGenerator.java
m-linner-ericsson: The reason for multiple catches is to give more information. Why would you like just one?

a-breid-ericsson: If the thrown exceptions message is self explaining more information would not be needed. It reduces the number of catches. The code looks better.

m-linner-ericsson: That would require changes and digging in other places in the code and I don't want to do that in this pullrequest.

